PR #1, opened 2020-08-19
Support building test client on non-windows platforms

laudrup:
Fix CMake build script to link with openssl on non-windows platforms
and add some ugly ifdefs in the very primitive https_client to alias
the ssl namespace to either the asio::ssl namespace I'm trying to
mimick or my new namespace.

Also add an unused method enum and constructor arguement to the
context constructor to mirror the interface of the original
asio::ssl::context constructor.

This is to have some easy way of comparing results/behavior during
implementation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,47 +2,58 @@ cmake_minimum_required(VERSION 3.15)
 
 project(boost-asio-windows-sspi)
 
-set(Boost_USE_STATIC_LIBS ON)
-set(Boost_USE_MULTITHREADED ON)
-set(Boost_USE_STATIC_RUNTIME OFF)
-set(BOOST_ROOT "C:/boost_1_73_0/")
-set(BOOST_LIBRARYDIR "C:/boost_1_73_0/lib/")
+if (WIN32)
+  set(Boost_USE_STATIC_LIBS ON)
+  set(Boost_USE_MULTITHREADED ON)
+  set(Boost_USE_STATIC_RUNTIME OFF)
+  set(BOOST_ROOT "C:/boost_1_73_0/")
+  set(BOOST_LIBRARYDIR "C:/boost_1_73_0/lib/")
+endif()
 
 find_package(Boost REQUIRED COMPONENTS date_time regex)
 
 add_executable(https_client https_client.cpp)
 target_link_libraries(https_client PRIVATE
-    Boost::headers
-    Boost::date_time
-    Boost::regex
+  Boost::headers
+  Boost::date_time
+  Boost::regex
+  )
+
+if (WIN32)
+  target_link_libraries(https_client PRIVATE
     Crypt32
     Secur32
     )
 
-set(MSVC_WARNINGS
-      /W4 # Baseline reasonable warnings
-      /w14254 # 'operator': conversion from 'type1:field_bits' to 'type2:field_bits', possible loss of data
-      /w14263 # 'function': member function does not override any base class virtual member function
-      /w14265 # 'classname': class has virtual functions, but destructor is not virtual instances of this class may not
-              # be destructed correctly
-      /w14287 # 'operator': unsigned/negative constant mismatch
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -2,47 +2,58 @@ cmake_minimum_required(VERSION 3.15)
...
-set(Boost_USE_STATIC_LIBS ON)
-set(Boost_USE_MULTITHREADED ON)
-set(Boost_USE_STATIC_RUNTIME OFF)
-set(BOOST_ROOT "C:/boost_1_73_0/")
-set(BOOST_LIBRARYDIR "C:/boost_1_73_0/lib/")
+if (WIN32)
+  set(Boost_USE_STATIC_LIBS ON)
+  set(Boost_USE_MULTITHREADED ON)
+  set(Boost_USE_STATIC_RUNTIME OFF)
+  set(BOOST_ROOT "C:/boost_1_73_0/")
```
vinniefalco, line 9: umm.... :) probably best not to use a hard-coded directory. I believe `FindBoost` is the command, and it sets `$BOOST_ROOT` or something like that.

CMakeLists.txt
```diff
@@ -2,47 +2,58 @@ cmake_minimum_required(VERSION 3.15)
...
+    /w14547 # 'operator': operator before comma has no effect; expected operator with side-effect
+    /w14549 # 'operator': operator before comma has no effect; did you intend 'operator'?
+    /w14555 # expression has no effect; expected expression with side- effect
+    /w14640 # Enable warning on thread un-safe static member initialization
+    /w14826 # Conversion from 'type1' to 'type_2' is sign-extended. This may cause unexpected runtime behavior.
+    /w14905 # wide string literal cast to 'LPSTR'
+    /w14906 # string literal cast to 'LPWSTR'
+    /w14928 # illegal copy-initialization; more than one user-defined conversion has been implicitly applied
+    /permissive- # standards conformance mode for MSVC compiler.
+    /WX
```
vinniefalco, line 48: Warnings should be cranked up to full

https_client.cpp
```diff
@@ -1,14 +1,27 @@
...
 #include <iterator>
 #include <string>
 #include <fstream>
 
+#ifdef _WIN32
+namespace ssl = boost::asio::windows_sspi;
+#else
+namespace ssl = boost::asio::ssl;
+#endif
+
```
vinniefalco, line 24: Consider using
```
namespace net = boost::asio;
```

CMakeLists.txt
laudrup: Indeed it isn't, but in order to use `FindBoost` which is what the `FindPackage` command used just below does, it has to know the boost root directory.

I'm far from being an expert in CMake, but I believe it uses `pkg-config` on platforms where that is supported whereas you have to give it some place to look on Windows which is why I'm setting the root manually.

Anyway, it would most likely make sense to remove this specific line and just make the user have to specify that on the command line. Bare in mind that the user was only me and I didn't really plan for this file to stay :-)

CMakeLists.txt
laudrup: I'm fairly certain that's what I'm doing. I should probably give credit to Jason Turner since I took it from [here](https://github.com/lefticus/cpp_starter_project).

https_client.cpp
laudrup: Worth considering. I'll also consider removing `using boost::asio::ip::tcp;` since I personally hate pulling in symbols without a qualified name, but these things can most likely be considered related.

CMakeLists.txt
madmongo1: Yes, the normal thing in CMake is that if the user's boost is in a custom location (for example when cross-compiling) they would specify: `cmake -DBOOT_ROOT=<my_custom_directory> .` on the first invocation of CMake. The FindBoost.cmake script will know what to do from there.

https://cmake.org/cmake/help/latest/module/FindBoost.html

PR #2, opened 2020-08-30
Async write

laudrup:
Attempt at making write operations asynchronous. This also helps taking steps towards an actual design.

### DIFF
--- a/https_client.cpp
+++ b/https_client.cpp
@@ -73,8 +73,17 @@ int main(int argc, char *argv[]) {
     request_stream << "Accept: */*\r\n";
     request_stream << "Connection: close\r\n\r\n";
 
+    auto size_to_send = request.size();
+    boost::asio::async_write(socket, request,
+        [&size_to_send](const boost::system::error_code& error, std::size_t length) {
+          if (error || length != size_to_send) {
+            // TODO: Handle in a better way at some point
+            std::cerr << "Error sending\n";
+            abort();
+          }
+        });
     // Send the request.
-    net::write(socket, request);
+    io_ctx.run_one();
 
     // Read the response status line. The response streambuf will automatically
     // grow to accommodate the entire line. The growth may be limited by passing
--- a/include/boost/asio/windows_sspi/detail/sspi_functions.hpp
+++ b/include/boost/asio/windows_sspi/detail/sspi_functions.hpp
@@ -1,0 +1,123 @@
+//
+// windows_sspi/detail/sspi_functions.hpp
+// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+//
+// Copyright (c) 2020 Kasper Laudrup (laudrup at stacktrace dot dk)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+
+#ifndef BOOST_ASIO_WINDOWS_SSPI_DETAIL_SSPI_FUNCTIONS_HPP
+#define BOOST_ASIO_WINDOWS_SSPI_DETAIL_SSPI_FUNCTIONS_HPP
+
+// TODO: Avoid cluttering global namespace (and avoid Windows headers if possible)
+#define SECURITY_WIN32
+#include <security.h>
+
+namespace boost {
+namespace asio {
+namespace windows_sspi {
+namespace detail {
+namespace sspi_functions {
+
+inline SecurityFunctionTable* sspi_function_table() {
+  static SecurityFunctionTable* impl = InitSecurityInterface();
+  // TODO: Figure out some way to signal this to the user instead of aborting
+  BOOST_ASSERT_MSG(impl != nullptr, "Unable to initialize SecurityFunctionTable");
+  return impl;
+}
+
+inline SECURITY_STATUS AcquireCredentialsHandleA(LPSTR pszPrincipal,
+                                                 LPSTR pszPackage,
+                                                 unsigned long fCredentialUse,
+                                                 void* pvLogonId,
+                                                 void* pAuthData,
+                                                 SEC_GET_KEY_FN pGetKeyFn,
+                                                 void* pvGetKeyArgument,
+                                                 PCredHandle phCredential,
+                                                 PTimeStamp ptsExpiry) {
+  return sspi_function_table()->AcquireCredentialsHandleA(pszPrincipal,
+                                                          pszPackage,
+                                                          fCredentialUse,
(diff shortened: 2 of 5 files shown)


https_client.cpp
```diff
@@ -73,8 +73,17 @@ int main(int argc, char *argv[]) {
     request_stream << "Accept: */*\r\n";
     request_stream << "Connection: close\r\n\r\n";
 
+    auto size_to_send = request.size();
+    boost::asio::async_write(socket, request,
```
vinniefalco, line 77: I like making a namespace alias `net` for `boost::asio` and using that for all APIs that map to the Networking TS

include/boost/asio/windows_sspi/stream.hpp
```diff
@@ -92,10 +98,51 @@ class sspi_impl {
   CtxtHandle* m_context;
 };
 
+// TODO: Move away from this file
+template <typename NextLayer, typename ConstBufferSequence>
+struct async_write_impl {
+  async_write_impl(NextLayer& next_layer, const ConstBufferSequence& buffer, std::shared_ptr<sspi_impl> sspi)
+    : m_next_layer(next_layer)
+    , m_buffer(buffer)
+    , m_sspi(sspi)
```
vinniefalco, line 107: `std::move(sspi)` to avoid an unnecessary atomic operation

include/boost/asio/windows_sspi/stream.hpp
```diff
@@ -92,10 +98,51 @@ class sspi_impl {
...
+    , m_state(encrypt) {
+  }
+
+  template <typename Self>
+  void operator()(Self& self, boost::system::error_code ec = {}, std::size_t length = 0) {
+    boost::ignore_unused(length);
+    switch (m_state) {
```
vinniefalco, line 114: If you use Asio's fauxroutine you can express this code more naturally, instead of the switch statement. It will be easier to read and maintain.

include/boost/asio/windows_sspi/stream.hpp
```diff
@@ -332,25 +328,22 @@ template <typename NextLayer> class stream : public stream_base {
...
   }
 
   template <typename ConstBufferSequence, typename CompletionToken>
   auto async_write_some(const ConstBufferSequence &buffer, CompletionToken &&token) ->
-      typename boost::asio::async_result<typename std::decay<CompletionToken>::type,
-                                         void(boost::system::error_code, std::size_t)>::return_type {
-    return boost::asio::async_compose<CompletionToken, void(boost::system::error_code, std::size_t)>(
+      typename net::async_result<typename std::decay<CompletionToken>::type,
+                                 void(boost::system::error_code, std::size_t)>::return_type {
+    return net::async_compose<CompletionToken, void(boost::system::error_code, std::size_t)>(
```
vinniefalco, line 346: `async_compose` is not part of Networking TS so this should just read `boost::asio::async_compose`

include/boost/asio/windows_sspi/stream.hpp
laudrup: Good point. Thanks.

include/boost/asio/windows_sspi/stream.hpp
laudrup: Thanks. I'll look into starting with this helps me get an understanding of what is going on though.

include/boost/asio/windows_sspi/stream.hpp
laudrup: I see. So the main reason why you prefer aliasing `boost::asio` to `net` is because of Networking TS?

Hadn't thought of that, but I guess that's similar to `namespace fs = boost::filesystem`.

I'll change that.

vinniefalco: Three reasons for `net`:
1. To abstract the difference between standalone Asio and Boost.Asio
2. To indicate to the reader that the function or type is a candidate for standardization
3. Because it is shorter (and thus easier to read)

It doesn't affect the correctness, but I place value on being informative. It can help reviewers and people who are interested in learning. It also helps me keep tabs on where the differences are between Asio and the upcoming networking standard. It isn't strictly necessary.

laudrup: I definitely agree with the third point, but just hadn't considered the two other points. It definitely makes sense.

I find it important that code is easy to read. It's hard to convince yourself that the code is correct if you have a hard time reading it in the first place, so I definitely find things like these important as well.

include/boost/asio/windows_sspi/detail/sspi_functions.hpp
```diff
@@ -0,0 +1,127 @@
...
+//
+
+#ifndef BOOST_ASIO_WINDOWS_SSPI_DETAIL_SSPI_FUNCTIONS_HPP
+#define BOOST_ASIO_WINDOWS_SSPI_DETAIL_SSPI_FUNCTIONS_HPP
+
+// TODO: Avoid cluttering global namespace (and avoid Windows headers if possible)
+#define SECURITY_WIN32
+#include <security.h>
+
+namespace {
```
djarek, line 18: Unnamed namespaces in headers are a recipe for ODR violations, because functions defined with non-internal linkage will have a different definition of the entities in that namespace in each TU that includes this header.

include/boost/asio/windows_sspi/detail/sspi_functions.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+} // namespace
+
+namespace boost {
+namespace asio {
+namespace windows_sspi {
+namespace detail {
+namespace sspi_functions {
+
+SECURITY_STATUS AcquireCredentialsHandleA(LPSTR pszPrincipal,
```
djarek, line 35: Free-functions defined in headers need to be `inline` to avoid ODR issues.

include/boost/asio/windows_sspi/detail/sspi_functions.hpp
laudrup: Of course. Thanks for pointing this out. Guess it would be OK to move this out of the anonymous namespace? It is still in a `detail` namespace after all?

include/boost/asio/windows_sspi/detail/sspi_functions.hpp
laudrup: Indeed they should. This is kind of embarrassing :-(

include/boost/asio/windows_sspi/detail/sspi_functions.hpp
vinniefalco: You can assume that users will never use identifiers in any `detail` namespace (no matter where it is nested in your headers.

PR #3, opened 2020-09-08
Async read

laudrup:
Implement async read similar to how async write has been implemented.

Still need to come up with a better design/separation between the SSPI encryption/decryption code and the public interface, but this is a small step in that direction.

Apart from that tons of things still needs to be implemented, but first step for now would probably be to look into some way of unittesting so I at least ensure I don't break things more than they already are.

### DIFF
--- a/https_client.cpp
+++ b/https_client.cpp
@@ -1,20 +1,20 @@
 #ifdef _WIN32
-# include <SDKDDKVer.h>
+#include <SDKDDKVer.h>
 #endif
 
 #include <boost/asio.hpp>
 
 #ifdef _WIN32
-# include <boost/asio/windows_sspi.hpp>
+#include <boost/asio/windows_sspi.hpp>
 #else
-# include <boost/asio/ssl.hpp>
+#include <boost/asio/ssl.hpp>
 #endif
 
 #include <cstdlib>
+#include <fstream>
 #include <iostream>
 #include <iterator>
 #include <string>
-#include <fstream>
 
 namespace net = boost::asio;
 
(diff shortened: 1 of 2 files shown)


https_client.cpp
```diff
@@ -1,20 +1,20 @@
 #ifdef _WIN32
-# include <SDKDDKVer.h>
+#include <SDKDDKVer.h>
 #endif
 
 #include <boost/asio.hpp>
 
 #ifdef _WIN32
-# include <boost/asio/windows_sspi.hpp>
+#include <boost/asio/windows_sspi.hpp>
```
vinniefalco, line 8: you really should use your own directory for these files, not boost/asio. this will avoid problems later

include/boost/asio/windows_sspi/stream.hpp
```diff
@@ -119,14 +162,67 @@ template <typename NextLayer, typename ConstBufferSequence> struct async_write_i
...
+    , m_sspi_impl(std::move(sspi_impl)) {
+  }
+
+  template <typename Self> void operator()(Self& self, boost::system::error_code ec = {}, std::size_t length = 0) {
+    BOOST_ASIO_CORO_REENTER(*this) {
+      while(m_sspi_impl->decrypted_data.empty()) {
+        // TODO: Find some way to make the sspi_impl, the decrypt
+        // function or something else be responsible for keeping track
+        // of state and buffer(s)
+        m_message.resize(0x10000 - m_sspi_impl->encrypted_data.size());
```
vinniefalco, line 191: Consider an assert to make sure the calculation does not underflow

include/boost/asio/windows_sspi/stream.hpp
```diff
@@ -119,14 +162,67 @@ template <typename NextLayer, typename ConstBufferSequence> struct async_write_i
...
+          return;
+        }
+        m_message.resize(length);
+        auto sc = m_sspi_impl->decrypt(m_message);
+        m_message.clear();
+        if (sc == SEC_E_INCOMPLETE_MESSAGE) {
+          continue;
+        }
+        if (sc == SEC_I_CONTEXT_EXPIRED) {
+          self.complete(net::error::eof, 0);
```
vinniefalco, line 206: is this really EOF though?

include/boost/asio/windows_sspi/stream.hpp
```diff
@@ -119,14 +162,67 @@ template <typename NextLayer, typename ConstBufferSequence> struct async_write_i
...
+          return;
+        }
+        if (sc != SEC_E_OK) {
+          self.complete(error::make_error_code(sc), 0);
+          return;
+        }
+      }
+
+      std::size_t to_return = std::min(net::buffer_size(m_buffer), m_sspi_impl->decrypted_data.size());
+      net::buffer_copy(m_buffer, net::buffer(m_sspi_impl->decrypted_data, to_return));
```
vinniefalco, line 216: `buffer_copy` returns the number of bytes actually copied. Consider asserting that this number is what you expected.

laudrup: Good point. Guess it should be an assertion as it would be a programming error if I somehow messed up the call to `std::min` above (or whatever this might look like in the future).

https_client.cpp
laudrup: Will do. I just copied this from the [gnutls implementation](https://github.com/paullouisageneau/boost-asio-gnutls) that I originally tried to mimic.

Should I just use something like `boost-asio-winsspi` as the root of the project/include dir?

include/boost/asio/windows_sspi/stream.hpp
laudrup: Good point. That's actually a real risk in this case. Of course, I'd rather handle all of this differently, but that doesn't make the overflow less of an issue currently.

https_client.cpp
vinniefalco: I would use
```
#include <boost/windows_sspi/windows_sspi.hpp>
```

In other words, the name of your library is Boost.WindowsSSPI

include/boost/asio/windows_sspi/stream.hpp
laudrup: [`The DecryptMessage (Schannel) function returns SEC_I_CONTEXT_EXPIRED when the message sender has shut down the connection.`](https://docs.microsoft.com/en-us/windows/win32/secauthn/shutting-down-an-schannel-connection).

I guess that's best handled by returning EOF to the client? Is there some better error code to use?

The same documentation also states that this should be followed by shutting down the context (which has not yet been implemented). The openssl implementation requires the user to do that though, so it might at least be a good idea for me to look into this.

vinniefalco: I'm not sure what the right answer is. At some point you will have to define your own error codes, if anything to deliver "short read". Or I guess you could deliver the corresponding asio error codes.

laudrup: Looking more into it, I was wrong in that the openssl implementation requires the user to do anything.

The correct solution is to shut down the connection gracefully (by following the documentation from Microsoft) and *then* I think it would be OK to send EOF to the client.

The whole async shutdown thing needs to be implemented though, but in short, you are absolutely right that simply returning EOF here is not correct.

https_client.cpp
laudrup: OK. Thanks. I'll do a `git move` :-)

PR #4, opened 2020-09-09
Move implementation out of asio namespace/directory

laudrup:
This implementation is not part of boost::asio so move it out of the
boost::asio namespace and directory.

It is intended to be a boost library, at least for now by name, so
keep it in the boost namespace and directory.


https_client.cpp
```diff
@@ -19,7 +19,7 @@
 namespace net = boost::asio;
 
 #ifdef _WIN32
-namespace ssl = net::windows_sspi;
+namespace ssl = boost::windows_sspi;
 #else
 namespace ssl = net::ssl;
```
vinniefalco, line 24: There is no `net::ssl` in the Networking TS btw

include/boost/windows_sspi/context.hpp
```diff
@@ -8,12 +8,12 @@
...
 
-#include <boost/asio/windows_sspi/context_base.hpp>
-#include <boost/asio/windows_sspi/detail/sspi_functions.hpp>
-#include <boost/asio/windows_sspi/error.hpp>
+#include <boost/windows_sspi/context_base.hpp>
+#include <boost/windows_sspi/detail/sspi_functions.hpp>
+#include <boost/windows_sspi/error.hpp>
 
 // TODO: Avoid cluttering global namespace (and avoid Windows headers if possible)
 #define SECURITY_WIN32
```
vinniefalco, line 19: macros belonging to your library need to be prefixed with `BOOST_WINDOWS_SSPI_`

https_client.cpp
laudrup: I guess I should use `namespace ssl = boost::asio::ssl` instead then?

include/boost/windows_sspi/context.hpp
laudrup: If I had to define a macro, I would try to avoid doing so in a header, but if I had to define a macro in a header, I would do my best to ensure that it wouldn't effect users of header and even if I felt there was no way to avoid that, I definitely wouldn't name it something this generic :-)

Anyway, this specific macro needs to be defined before including the relevant Windows headers for SSPI. The TODO is there to remind me to try and avoid including the Windows headers as they are most often very nasty and can break tons of completely unrelated code, which I of course want to avoid.

If there ever was a award for worst C header ever written, I would be surprised if `windows.h` wouldn't win that award :-)

https_client.cpp
vinniefalco: That would be ok

include/boost/windows_sspi/context.hpp
vinniefalco: See https://github.com/boostorg/winapi

laudrup: > See https://github.com/boostorg/winapi

Thanks, I've looked into that already. There's a reason that exists and I might be able to use some parts of it, but unfortunately it doesn't really solve all the issues.

As an example, I would probably be able to map the error codes returned from SSPI into an enum an use that instead without too many worries, but I'm a bit more worried about redefining a struct that is returned from one of the Windows API functions as a pointer (or passed to the function as a pointer) and assume that I could just use that. I would assume that could cause some really nasty and hard to debug issues for potential users.

I have looked a bit into how other libraries (including boost libraries) are dealing with the pain of supporting Windows, but I haven't really found any silver bullet.

Ensuring that I don't define this macro (`WIN32_SECURITY`) for any users of my headers would probably be a huge step in the right direction though. It seems to mean that I want to use SSPI in a user mode application (as opposed to a kernel driver) btw.

vinniefalco: Yeah Boost.WinApi is not perfect. And it is of course going to be missing most of the very-specific APIs.

PR #5, opened 2020-09-13
Unittest

laudrup:


CMakeLists.txt
```diff
@@ -1,30 +1,33 @@
 cmake_minimum_required(VERSION 3.15)
 
-project(boost-asio-windows-sspi CXX)
+project(boost-windows-sspi CXX)
 
 set(CMAKE_CXX_STANDARD 11)
 
+enable_testing()
```
djarek, line 7: `enable_testing` enables tests unconditionally, however, you might want to use the standard CMake approach:
```
include(CTest)
if(BUILD_TESTING)
    enable_testing()
    add_subdirectory(tests)
endif()
```

Which allows users to disable building tests of your library. This is useful for users consuming the library e.g. via `vcpkg`, so that they don't need to distribute your test binaries in their `vcpkg export`.

CMakeLists.txt
```diff
@@ -1,30 +1,33 @@
...
-add_executable(https_client https_client.cpp)
-target_link_libraries(https_client PRIVATE
-  Boost::headers
-  Boost::date_time
-  Boost::regex
-  )
+  add_library(boost-windows-sspi INTERFACE)
+  target_link_libraries(boost-windows-sspi INTERFACE
+    Boost::headers
+    Boost::date_time
```
djarek, line 19: Are you sure linking `date_time` and `regex` is necessary? I thought they're optional transitive dependencies.

examples/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+add_executable(https_client https_client.cpp)
+
+if (WIN32)
+  target_link_libraries(https_client PRIVATE
+    boost-windows-sspi
+    )
+else()
+  target_link_libraries(https_client PRIVATE
+    Boost::headers
+    pthread
```
djarek, line 10: The idiomatic way of enabling threading is:
```
find_package (Threads)

target_link_libraries(https_client PRIVATE Threads::Threads)
```

CMakeLists.txt
laudrup: Good point. I have considered making the build more customizable in different ways, but I might as well look into that right now. Thanks.

CMakeLists.txt
laudrup: To be honest, I can't really remember why they were necessary and what depended on them. I'll look into that.

examples/CMakeLists.txt
laudrup: Indeed. I thought I had actually done that. Will do. Thanks.

PR #7, opened 2020-09-22
Extend unittest to make client read response

laudrup:
This is a few commits so the branch name is not really descriptive.

Apart from extending the unit test to actually test that the data being sent is the same as being received, this is an attempt towards an actual design where the decryption and encryption is handled by classes that manages state and buffers.

The next thing to do is to extend the tests to test the async implementation and then make the handshake async or implement async handshake and then test it. Whatever makes most sense.

I would really like some feedback on whether the design is moving in the right direction.

### DIFF
--- a/include/boost/windows_sspi/detail/sspi_impl.hpp
+++ b/include/boost/windows_sspi/detail/sspi_impl.hpp
@@ -21,25 +21,14 @@ namespace detail {
 
 namespace net = boost::asio;
 
-class sspi_impl {
+class sspi_encrypt {
 public:
-  sspi_impl(CtxtHandle* context)
+  sspi_encrypt(CtxtHandle* context)
     : m_context(context) {
   }
 
-  // TODO: Calculate this once when handshake is complete
-  SecPkgContext_StreamSizes stream_sizes() const {
-    SecPkgContext_StreamSizes stream_sizes;
-    SECURITY_STATUS sc = detail::sspi_functions::QueryContextAttributes(m_context, SECPKG_ATTR_STREAM_SIZES, &stream_sizes);
-
-    // TODO: Signal error to user (throw exception or use error code?)
-    boost::ignore_unused(sc);
-    BOOST_ASSERT(sc == SEC_E_OK);
-    return stream_sizes;
-  }
-
   template <typename ConstBufferSequence>
-  std::vector<char> encrypt(const ConstBufferSequence& buffers, boost::system::error_code& ec, size_t& size_encrypted) {
+  std::vector<char> operator()(const ConstBufferSequence& buffers, boost::system::error_code& ec, size_t& size_encrypted) {
     SecBufferDesc Message;
     SecBuffer Buffers[4];
 
@@ -76,7 +65,28 @@ public:
     return message;
   }
 
-  SECURITY_STATUS decrypt(const std::vector<char>& data) {
+private:
+  // TODO: Calculate this once when handshake is complete
+  SecPkgContext_StreamSizes stream_sizes() const {
+    SecPkgContext_StreamSizes stream_sizes;
+    SECURITY_STATUS sc = detail::sspi_functions::QueryContextAttributes(m_context, SECPKG_ATTR_STREAM_SIZES, &stream_sizes);
+
+    // TODO: Signal error to user (throw exception or use error code?)
+    boost::ignore_unused(sc);
+    BOOST_ASSERT(sc == SEC_E_OK);
+    return stream_sizes;
+  }
+
+  CtxtHandle* m_context;
+};
+
+class sspi_decrypt {
+public:
+  sspi_decrypt(CtxtHandle* context)
+    : m_context(context) {
+  }
+
+  SECURITY_STATUS operator()(const std::vector<char>& data) {
     encrypted_data.insert(encrypted_data.end(), data.begin(), data.end());
 
     SecBufferDesc Message;
@@ -117,6 +127,17 @@ public:
 
 private:
   CtxtHandle* m_context;
+};
+
+class sspi_impl {
+public:
+  sspi_impl(CtxtHandle* context)
+    : encrypt(context)
+    , decrypt(context) {
+  }
+
+  sspi_encrypt encrypt;
+  sspi_decrypt decrypt;
 };
 
 } // namespace detail
--- a/include/boost/windows_sspi/detail/sspi_impl.hpp
+++ b/include/boost/windows_sspi/detail/sspi_impl.hpp
@@ -12,14 +12,13 @@
 #define BOOST_WINDOWS_SSPI_DETAIL_SSPI_IMPL_HPP
 
 #include <boost/windows_sspi/detail/sspi_functions.hpp>
+#include <boost/windows_sspi/detail/config.hpp>
 
 #include <boost/core/ignore_unused.hpp>
 
 namespace boost {
 namespace windows_sspi {
 namespace detail {
-
-namespace net = boost::asio;
 
 class sspi_encrypt {
 public:
--- a/include/boost/windows_sspi/detail/sspi_impl.hpp
+++ b/include/boost/windows_sspi/detail/sspi_impl.hpp
@@ -27,6 +27,7 @@ public:
     : m_context(context) {
   }
 
+  // TODO: Since this is now a functor class, consider getting rid of these output arguments
   template <typename ConstBufferSequence>
   std::vector<char> operator()(const ConstBufferSequence& buffers, boost::system::error_code& ec, size_t& size_encrypted) {
     SecBufferDesc Message;
@@ -83,11 +84,23 @@ private:
 class sspi_decrypt {
 public:
   sspi_decrypt(CtxtHandle* context)
-    : m_context(context) {
+    : error_code(SEC_E_OK)
+    , m_context(context) {
   }
 
-  SECURITY_STATUS operator()(const std::vector<char>& data) {
-    encrypted_data.insert(encrypted_data.end(), data.begin(), data.end());
+  enum class state {
+    data_needed,
+    data_available,
+    error
+  };
+
+  state operator()() {
+    if (!decrypted_data.empty()) {
+      return state::data_available;
+    }
+    if (encrypted_data.empty()) {
+      return state::data_needed;
+    }
 
     SecBufferDesc Message;
     SecBuffer Buffers[4];
@@ -103,9 +116,12 @@ public:
     Message.cBuffers = 4;
     Message.pBuffers = Buffers;
 
-    SECURITY_STATUS sc = detail::sspi_functions::DecryptMessage(m_context, &Message, 0, NULL);
-    if (sc != SEC_E_OK) {
-      return sc;
+    error_code = detail::sspi_functions::DecryptMessage(m_context, &Message, 0, NULL);
+    if (error_code == SEC_E_INCOMPLETE_MESSAGE) {
+      return state::data_needed;
+    }
+    if (error_code != SEC_E_OK) {
+      return state::error;
     }
 
     encrypted_data.clear();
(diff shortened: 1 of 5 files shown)


madmongo1:
Hi @laudrup ,

Vinnie has asked me to review, would you mind requesting a review from me?

laudrup:
> Hi @laudrup ,
> 
> Vinnie has asked me to review, would you mind requesting a review from me?

Of course. I'd be more than happy to.

I guess you need to be a collaborator first?

Thanks a lot!

madmongo1:
Is this something you can do from the project settings or do I need to submit a PR?

laudrup:
> Is this something you can do from the project settings or do I need to submit a PR?

You should have received an invitation now.

include/boost/windows_sspi/stream.hpp
```diff
@@ -109,11 +109,11 @@ template <typename NextLayer, typename MutableBufferSequence> struct async_read_
...
       }
 
-      std::size_t to_return = std::min(net::buffer_size(m_buffer), m_sspi_impl->decrypted_data.size());
-      std::size_t bytes_copied = net::buffer_copy(m_buffer, net::buffer(m_sspi_impl->decrypted_data, to_return));
+      std::size_t to_return = std::min(net::buffer_size(m_buffer), m_sspi_impl->decrypt.decrypted_data.size());
+      std::size_t bytes_copied = net::buffer_copy(m_buffer, net::buffer(m_sspi_impl->decrypt.decrypted_data, to_return));
       boost::ignore_unused(bytes_copied);
       BOOST_ASSERT(bytes_copied == to_return);
-      m_sspi_impl->decrypted_data.erase(m_sspi_impl->decrypted_data.begin(), m_sspi_impl->decrypted_data.begin() + to_return);
+      m_sspi_impl->decrypt.decrypted_data.erase(m_sspi_impl->decrypt.decrypted_data.begin(), m_sspi_impl->decrypt.decrypted_data.begin() + to_return);
```
madmongo1: It makes me a little sad to see a vector erase operation in a loop - it's probably immaterial in practice, but theoretically I wonder if there's a cleaner way (circular buffer, segmented buffer, decrypt directly into client buffer?)

include/boost/windows_sspi/detail/sspi_impl.hpp
```diff
@@ -119,11 +135,28 @@ class sspi_decrypt {
...
+    std::size_t size = std::min(max, decrypted_data.size());
+    std::vector<char> ret{decrypted_data.begin(), decrypted_data.begin() + size};
+    decrypted_data.erase(decrypted_data.begin(), decrypted_data.begin() + size);
+    return ret;
   }
 
+  // TODO: Make private
   std::vector<char> encrypted_data;
   std::vector<char> decrypted_data;
+  SECURITY_STATUS error_code;
```
madmongo1, line 159: naming this variable error_code threw me off since it's the name name as the system::error_code type.

include/boost/windows_sspi/detail/sspi_impl.hpp
```diff
@@ -119,11 +135,28 @@ class sspi_decrypt {
...
+    return state::data_available;
+  }
+
+  // TODO: Consider making this more flexible by not requering a
+  // vector of chars, but any view of a range of bytes
+  void put(const std::vector<char>& data) {
+    encrypted_data.insert(encrypted_data.end(), data.begin(), data.end());
+  }
+
+  std::vector<char> get(std::size_t max) {
```
madmongo1, line 149: returning a new vector here forces memory allocation. I wonder whether there is room for another version of this function that takes a reference to a vector to modify? This way, the caller is able to avoid the pessimisation by re-using the same vector on each call.

include/boost/windows_sspi/detail/sspi_impl.hpp
```diff
@@ -119,11 +135,28 @@ class sspi_decrypt {
...
+    std::size_t size = std::min(max, decrypted_data.size());
+    std::vector<char> ret{decrypted_data.begin(), decrypted_data.begin() + size};
+    decrypted_data.erase(decrypted_data.begin(), decrypted_data.begin() + size);
```
madmongo1, line 152: If called in a loop, I wonder if there is a better way than erase()?

include/boost/windows_sspi/stream.hpp
```diff
@@ -68,61 +68,43 @@ template <typename NextLayer, typename ConstBufferSequence> struct async_write_i
...
-        // TODO: Fix so overflow cannot happen (we don't need to read
-        // more data unless DecryptMessage asks us to).
-        BOOST_ASSERT(m_sspi_impl->decrypt.encrypted_data.size() < 0x10000);
-        m_message.resize(0x10000 - m_sspi_impl->decrypt.encrypted_data.size());
+      while(m_sspi_impl->decrypt() == detail::sspi_decrypt::state::data_needed) {
+        // TODO: Use a fixed size buffer instead
+        m_input.resize(0x10000);
         BOOST_ASIO_CORO_YIELD net::async_read(m_next_layer,
-                                              net::buffer(m_message),
+                                              net::buffer(m_input),
```
madmongo1, line 83: async_read will 'hang' until 64k bytes read or end of stream. Is this the intention? The inititation function is async_read_some, which a user would expect to complete quickly if there was already any data in the tcp receive window.

You may want to consider async_read_some - which will be just as efficient with a large buffer.

include/boost/windows_sspi/stream.hpp
```diff
@@ -68,61 +68,43 @@ template <typename NextLayer, typename ConstBufferSequence> struct async_write_i
...
-      m_sspi_impl->decrypt.decrypted_data.erase(m_sspi_impl->decrypt.decrypted_data.begin(), m_sspi_impl->decrypt.decrypted_data.begin() + to_return);
-      self.complete(boost::system::error_code{}, to_return);
+      if (m_sspi_impl->decrypt() == detail::sspi_decrypt::state::error) {
+        ec = boost::error::make_error_code(m_sspi_impl->decrypt.error_code);
+        self.complete(ec, 0);
+        return;
+      }
+
+      const auto data = m_sspi_impl->decrypt.get(net::buffer_size(m_buffers));
+      std::size_t bytes_copied = net::buffer_copy(m_buffers, net::buffer(data));
```
madmongo1, line 97: I can't help thinking that decrypting directly into user buffers will avoid a redundant copy.

include/boost/windows_sspi/stream.hpp
```diff
@@ -275,44 +257,22 @@ template <typename NextLayer> class stream : public stream_base {
...
-        return 0;
-      }
-      if (sc != SEC_E_OK) {
-        ec = error::make_error_code(sc);
-        return 0;
-      }
-      break;
+    while(m_sspi_impl->decrypt() == detail::sspi_decrypt::state::data_needed) {
+      std::array<char, 0x10000> input_buffer;
+      std::size_t size_read = m_next_layer.read_some(net::buffer(input_buffer.data(), input_buffer.size()), ec);
```
madmongo1, line 262: read_some used here whereas the async version used async_read[!_some] - I believe the sync version is correct.

include/boost/windows_sspi/stream.hpp
```diff
@@ -275,44 +257,22 @@ template <typename NextLayer> class stream : public stream_base {
...
+      m_sspi_impl->decrypt.put({input_buffer.begin(), input_buffer.begin() + size_read});
+      continue;
+    }
+
+    if (m_sspi_impl->decrypt() == detail::sspi_decrypt::state::error) {
+      ec = boost::error::make_error_code(m_sspi_impl->decrypt.error_code);
```
madmongo1, line 268: `make_error_code` is designed to be found through ADL. This doesn't work for you because `SECURITY_STATUS` is not an enum.
If it were, you could create a specialilsation of `boost::system::is_error_code_enum<SECURITY_STATUS> : boost::true_type {};` and then simply write `ec = m_sspi_impl->decrypt.error_code;`

madmongo1 (changes requested):
all but one comments are QOI queries. The only serious one is the use of `async_read` rather than `async_read_some`.

include/boost/windows_sspi/detail/sspi_impl.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+#include <boost/windows_sspi/detail/sspi_functions.hpp>
+
+#include <boost/core/ignore_unused.hpp>
+
+namespace boost {
+namespace windows_sspi {
+namespace detail {
+
+namespace net = boost::asio;
```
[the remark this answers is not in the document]
laudrup: Makes sense. Will do.

include/boost/windows_sspi/detail/sspi_impl.hpp
```diff
@@ -0,0 +1,126 @@
...
+  sspi_impl(CtxtHandle* context)
+    : m_context(context) {
+  }
+
+  // TODO: Calculate this once when handshake is complete
+  SecPkgContext_StreamSizes stream_sizes() const {
+    SecPkgContext_StreamSizes stream_sizes;
+    SECURITY_STATUS sc = detail::sspi_functions::QueryContextAttributes(m_context, SECPKG_ATTR_STREAM_SIZES, &stream_sizes);
+
+    // TODO: Signal error to user (throw exception or use error code?)
```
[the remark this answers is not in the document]
laudrup: The most likely way that this would fail would be if the handshake hasn't been completed, which would most likely give a rather cryptic error message.

It would most likely be better to ensure that calling any of the read or write functions would fail with a useful error if that was the case, but for now I'll refactor the code to ensure that an error is returned to the user. It is required if the function would fail for any other reason anyway.

include/boost/windows_sspi/stream.hpp
```diff
@@ -33,108 +34,9 @@ namespace windows_sspi {
...
-  std::vector<char> decrypted_data;
-
-private:
-  CtxtHandle* m_context;
-};
-
 // TODO: Move away from this file
 template <typename NextLayer, typename ConstBufferSequence> struct async_write_impl : boost::asio::coroutine {
-  async_write_impl(NextLayer& next_layer, const ConstBufferSequence& buffer, std::shared_ptr<sspi_impl> sspi_impl)
+  async_write_impl(NextLayer& next_layer, const ConstBufferSequence& buffer, std::shared_ptr<detail::sspi_impl> sspi_impl)
```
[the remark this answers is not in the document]
laudrup: Indeed. I don't really know why I made this a `shared_ptr` in the first place. I'll fix that.

include/boost/windows_sspi/detail/sspi_impl.hpp
laudrup: Understandable :-) I will renamed it.

include/boost/windows_sspi/stream.hpp
laudrup: That was definitely a mistake. Thanks for spotting this. Will fix.

include/boost/windows_sspi/stream.hpp
laudrup: I've been struggling a bit with making this work. Mostly because (I think) `SECURITY_STATUS` is actually just a `long` type, so I either the conversion to `error_code` already exists, or there's no conversion to a custom enum type if I declare that.

I'm sure I'll figure that out though, there are quite a few other boost libraries where I can find inspiration.

laudrup:
> all but one comments are QOI queries. The only serious one is the use of `async_read` rather than `async_read_some`.

That was definitely a mistake that was easy and important to fix. Thanks for finding that.

Most of the other comments are related to sub-optimal handling of buffers (e.g. unnecessary copies), which I definitely think is relevant, but I'd rather improve my test coverage a bit as well as refactoring the code a bit more before I start looking into these kind of improvements. I think it's more important to ensure that the code works first and ensure that there are proper tests of that before I start looking into performance improvements. I'm sure you agree.

I've added some TODOs on your points as that is currently my issue tracker, which I think is OK for a project in this early stage.

Thanks a lot for taking you time to do this. It means a lot.

include/boost/windows_sspi/stream.hpp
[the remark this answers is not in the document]
madmongo1: shared_ptr is an easy first choice when code is written off the cuff. It's often overused, but its selection is entirely understandable.

include/boost/windows_sspi/stream.hpp
madmongo1: the trick is to make it an enum and provide the boost::system plumbing (error_category, specialisation of is_enum_error_code<> and ADL make_error_code). If it's just copying values from the C api the enum could be empty. Then assignment becomes:
```
ec = decrypt.error_code;
```

madmongo1: Asio and beast have examples (boost/asio/error.hpp)

madmongo1:
My pleasure.

include/boost/windows_sspi/stream.hpp
laudrup: Thanks. I was a bit confused as to whether I could simply make it work without having to redeclare all the SECURITY_STATUS errors as enum values. Looking at the asio code, it does indeed seems like it possible, so thanks for clarifying that.

I'll look into that, but I'll probably just merge this PR as it is.

PR #8, opened 2020-10-04
Async handshake

laudrup:
This implements async handshake which means that most of the basic functionality has been implemented on the client side. The goal being to implement all of the functionality available in the existing `boost::asio::ssl` implementation.

I intend to extend the tests to test errors, short read/writes etc. which should thankfully be fairly easy to achieve using the convenient test stream from `boost::beast`. There are quite a few places where some buffer copies could be avoided and definitely quite a lot of other things that could and should be cleaned up, but I think that will be easier to do once I have extended the tests a bit.

I also need to look into implementing the server side functionality, which mainly means implementing server handshake as well as all the certificate handling.

### DIFF
--- a/examples/https_client.cpp
+++ b/examples/https_client.cpp
@@ -37,26 +37,28 @@ public:
     // TODO: Consider async connect
     net::connect(socket_.next_layer(), endpoint_iterator);
 
-    // TODO: Implement async handshake
-    boost::system::error_code ec;
-    socket_.handshake(ssl::stream_base::client, ec);
-
-    std::ostream request_stream(&request_);
-    request_stream << "GET " << path << " HTTP/1.0\r\n";
-    request_stream << "Host: " << host << "\r\n";
-    request_stream << "Accept: */*\r\n";
-    request_stream << "Connection: close\r\n\r\n";
-
-    boost::asio::async_write(socket_, request_, [this](const boost::system::error_code& ec, std::size_t) {
+    socket_.async_handshake(ssl::stream_base::client, [this, &path, &host](const boost::system::error_code& ec) {
       if (ec) {
-        std::cerr << "Error sending: request:" << ec.message() << "\n";
+        std::cerr << "Error handshaking:" << ec.message() << "\n";
         return;
       }
-      net::async_read_until(socket_, response_, "\r\n", [this](boost::system::error_code ec, size_t) {
+      std::ostream request_stream(&request_);
+      request_stream << "GET " << path << " HTTP/1.0\r\n";
+      request_stream << "Host: " << host << "\r\n";
+      request_stream << "Accept: */*\r\n";
+      request_stream << "Connection: close\r\n\r\n";
+
+      boost::asio::async_write(socket_, request_, [this](const boost::system::error_code& ec, std::size_t) {
         if (ec) {
-          std::cerr << "Error receiving response: " << ec.message() << "\n";
+          std::cerr << "Error sending request:" << ec.message() << "\n";
+          return;
         }
-        read_response();
+        net::async_read_until(socket_, response_, "\r\n", [this](boost::system::error_code ec, size_t) {
+          if (ec) {
+            std::cerr << "Error receiving response: " << ec.message() << "\n";
+          }
+          read_response();
+        });
       });
     });
   }
--- a/examples/https_client.cpp
+++ b/examples/https_client.cpp
@@ -38,7 +38,8 @@ public:
     net::connect(socket_.next_layer(), endpoint_iterator);
 
     // TODO: Implement async handshake
-    socket_.handshake(ssl::stream_base::client);
+    boost::system::error_code ec;
+    socket_.handshake(ssl::stream_base::client, ec);
 
     std::ostream request_stream(&request_);
     request_stream << "GET " << path << " HTTP/1.0\r\n";
--- a/examples/https_client.cpp
+++ b/examples/https_client.cpp
@@ -46,7 +46,6 @@ public:
       request_stream << "GET " << path << " HTTP/1.0\r\n";
       request_stream << "Host: " << host << "\r\n";
       request_stream << "Accept: */*\r\n";
-      request_stream << "Connection: close\r\n\r\n";
 
       boost::asio::async_write(socket_, request_, [this](const boost::system::error_code& ec, std::size_t) {
         if (ec) {
--- a/include/boost/windows_sspi/detail/sspi_impl.hpp
+++ b/include/boost/windows_sspi/detail/sspi_impl.hpp
@@ -380,6 +380,9 @@ public:
     , decrypt(&m_context) {
   }
 
+  sspi_impl(const sspi_impl&) = delete;
+  sspi_impl& operator=(const sspi_impl&) = delete;
+
   ~sspi_impl() {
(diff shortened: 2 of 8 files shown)


examples/https_client.cpp
```diff
@@ -37,25 +37,28 @@ class https_client {
...
-        std::cerr << "Error sending: request:" << ec.message() << "\n";
+        std::cerr << "Error handshaking:" << ec.message() << "\n";
         return;
       }
-      net::async_read_until(socket_, response_, "\r\n", [this](boost::system::error_code ec, size_t) {
+      std::ostream request_stream(&request_);
+      request_stream << "GET " << path << " HTTP/1.0\r\n";
+      request_stream << "Host: " << host << "\r\n";
+      request_stream << "Accept: */*\r\n";
+      request_stream << "Connection: close\r\n\r\n";
```
madmongo1, line 49: HTTP/1.0 implies `Connection: close`

include/boost/windows_sspi/detail/sspi_impl.hpp
```diff
@@ -17,11 +17,169 @@
...
+      case SEC_I_INCOMPLETE_CREDENTIALS:
+        BOOST_ASSERT_MSG(false, "client authentication not implemented");
+
+      default:
+        return state::error;
+    }
+  }
+
+  // TODO: Consider making this more flexible by not requering a
+  // vector of chars, but any view of a range of bytes
```
madmongo1, line 157: this would be a boost::span pre c++20, or a std::span >= c++20

madmongo1: or to be compatible with asio & beast, a `net::const_buffer`

include/boost/windows_sspi/detail/sspi_impl.hpp
```diff
@@ -203,20 +361,36 @@ class sspi_decrypt {
...
-    : encrypt(context)
-    , decrypt(context) {
+  sspi_impl(CredHandle* cred_handle)
+    : m_cred_handle(cred_handle)
+    , handshake(&m_context, cred_handle)
+    , encrypt(&m_context)
+    , decrypt(&m_context) {
+  }
+
+  ~sspi_impl() {
```
madmongo1, line 383: This class controls a resource and is (I think) copyable - you should probably disable the copy constructor and assignment operator.

include/boost/windows_sspi/stream.hpp
```diff
@@ -33,14 +33,67 @@ namespace boost {
...
+template <typename NextLayer>
+struct async_handshake_impl : boost::asio::coroutine {
+  async_handshake_impl(NextLayer& next_layer, detail::sspi_impl& sspi_impl)
+    : m_next_layer(next_layer)
+    , m_sspi_impl(sspi_impl) {
+  }
+
+  template <typename Self>
+  void operator()(Self& self, boost::system::error_code ec = {}, std::size_t length = 0) {
+    BOOST_ASIO_CORO_REENTER(*this) {
```
madmongo1, line 45: You can save yourself some error checking code repetition by putting this above the re-enter:
```
if (ec) 
  return self.complete(ec);
```

include/boost/windows_sspi/stream.hpp
```diff
@@ -33,14 +33,67 @@ namespace boost {
...
+        if (state == detail::sspi_handshake::state::data_available) {
+          m_output = m_sspi_impl.handshake.get();
+          BOOST_ASIO_CORO_YIELD net::async_write(m_next_layer, net::buffer(m_output), net::transfer_exactly(m_output.size()), std::move(self));
+          if (ec) {
+            self.complete(ec);
+            return;
+          }
+          continue;
+        }
+        if (state == detail::sspi_handshake::state::error) {
```
madmongo1, line 69: query: is it guaranteed that when you reach this line, at least one YIELD will have happened prior?
If not, you will cause the completion handler to execute on the wrong executor by calling self.complete(ec) directly here.
This can be mitigated by detecting that you are not in a continuation and:
```
if (!is_continuation_)
  BOOST_ASIO_CORO_YIELD 
    net::post(boost::beast::bind_front_handler(std::move(self), ec));
self.complete(ec);
```

include/boost/windows_sspi/stream.hpp
```diff
@@ -33,14 +33,67 @@ namespace boost {
...
+          self.complete(m_sspi_impl.handshake.last_error());
+          return;
+        }
+      }
+      BOOST_ASSERT(!m_sspi_impl.handshake.last_error());
+      self.complete(m_sspi_impl.handshake.last_error());
```
madmongo1, line 75: same question here.

include/boost/windows_sspi/stream.hpp
```diff
@@ -87,12 +143,13 @@ template <typename NextLayer, typename MutableBufferSequence> struct async_read_
         continue;
       }
 
-      if (m_sspi_impl.decrypt() == detail::sspi_decrypt::state::error) {
-        ec = boost::error::make_error_code(m_sspi_impl.decrypt.last_error);
+      if (state == detail::sspi_decrypt::state::error) {
+        ec = m_sspi_impl.decrypt.last_error();
         self.complete(ec, 0);
```
madmongo1, line 148: same question re completing on correct executor.

madmongo1 (commented):
Some questions to prompt some thoughts around correct completion handler invocation.

examples/https_client.cpp
laudrup: Indeed it does. This is in no way anywhere close to being a complete HTTP client, but I might as well remove this line.

I would like to rewrite this example to use `boost::beast` I think. And potentially use it to make some integration tests with some real servers.

include/boost/windows_sspi/detail/sspi_impl.hpp
laudrup: My thoughts as well. Probably a `net::const_buffer` would make most sense, but as I tried to indicate in my comment, I'd prefer to look into that after I have improved my testing. In my experience, it just makes changing things like this a lot easier.

include/boost/windows_sspi/detail/sspi_impl.hpp
laudrup: Whoops.

This should definitely not be copyable as the CtxHandle keeps some state related to the current connection/stream, so even if it is possible to implement a copy construction/assignment, it most likely wouldn't make any sense.

Nicely spotted, thanks.

include/boost/windows_sspi/stream.hpp
laudrup: That's a very good point. Thanks.

include/boost/windows_sspi/stream.hpp
laudrup: Even if it was guaranteed (and it isn't currently) I'd rather not rely on that being required to be guaranteed, as it could lead to (for me at least) very hard to debug errors.

I will look into it and can hopefully figure it out on my own, but is there some similar simple way to achieve this (I guess more or less "forcing" a yield) without using anything from `boost::beast`?

I don't think it makes too much sense to depend on `boost::beast` for this library, even if `boost::beast` is an amazing library of course :-)

PR #10, opened 2020-10-09
Shutdown

laudrup:
This implements the shutdown functionality which was the last thing missing on the client side in order to handle async client communication.

This makes it even more obvious that a lot of code is duplicated and should be deduplicated, but in order to do any of that, some more extensive should first be added.

In order to do that, I have switched to a more feature full test framework (catch2). For now that is mostly to try and focus on actually writing tests even if most of the functionality I use could most likely be written specifically for this library which seems to be what many other boost libraries seems to do.

### DIFF
--- a/include/boost/windows_sspi/detail/sspi_impl.hpp
+++ b/include/boost/windows_sspi/detail/sspi_impl.hpp
@@ -23,6 +23,14 @@ namespace boost {
 namespace windows_sspi {
 namespace detail {
 
+const DWORD context_flags =
+  ISC_REQ_SEQUENCE_DETECT | // Detect messages received out of sequence
+  ISC_REQ_REPLAY_DETECT | // Detect replayed messages
+  ISC_REQ_CONFIDENTIALITY | // Encrypt messages
+  ISC_RET_EXTENDED_ERROR | // When errors occur, the remote party will be notified
+  ISC_REQ_ALLOCATE_MEMORY | // Allocate buffers. Free them with FreeContextBuffer
+  ISC_REQ_STREAM; // Support a stream-oriented connection
+
 class sspi_handshake {
 public:
   enum class state {
@@ -35,10 +43,7 @@ public:
   sspi_handshake(CtxtHandle* context, CredHandle* credentials)
     : m_context(context)
     , m_credentials(credentials)
-    , m_last_error(SEC_E_OK)
-    , m_flags_out(0)
-    , m_flags_in(ISC_REQ_SEQUENCE_DETECT | ISC_REQ_REPLAY_DETECT | ISC_REQ_CONFIDENTIALITY |
-                 ISC_RET_EXTENDED_ERROR | ISC_REQ_ALLOCATE_MEMORY | ISC_REQ_STREAM) {
+    , m_last_error(SEC_E_OK) {
 
     SecBufferDesc OutBuffer;
     SecBuffer OutBuffers[1];
@@ -51,17 +56,19 @@ public:
     OutBuffer.pBuffers = OutBuffers;
     OutBuffer.ulVersion = SECBUFFER_VERSION;
 
+    DWORD out_flags = 0;
+
     m_last_error = detail::sspi_functions::InitializeSecurityContext(m_credentials,
                                                                      nullptr,
                                                                      nullptr,
-                                                                     m_flags_in,
+                                                                     context_flags,
                                                                      0,
                                                                      SECURITY_NATIVE_DREP,
                                                                      nullptr,
                                                                      0,
                                                                      m_context,
                                                                      &OutBuffer,
-                                                                     &m_flags_out,
+                                                                     &out_flags,
                                                                      nullptr);
     if (m_last_error == SEC_I_CONTINUE_NEEDED) {
       // TODO: Avoid this copy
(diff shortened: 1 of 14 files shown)


include/boost/windows_sspi/detail/sspi_impl.hpp
```diff
@@ -107,17 +114,19 @@ class sspi_handshake {
     OutBuffer.pBuffers = OutBuffers;
     OutBuffer.ulVersion = SECBUFFER_VERSION;
 
+    DWORD out_flags = 0;
+
     m_last_error = detail::sspi_functions::InitializeSecurityContext(m_credentials,
                                                                      m_context,
                                                                      nullptr,
-                                                                     m_flags_in,
+                                                                     context_flags,
```
djarek, line 122: Not an actionable item in the context of this PR: Might want to consider allowing a user to specify additional options in the future. IIRC, this is where you enable mTLS authentication.

include/boost/windows_sspi/stream.hpp
```diff
@@ -201,6 +200,53 @@ struct async_read_impl : boost::asio::coroutine {
...
+          BOOST_ASIO_CORO_YIELD net::async_write(m_next_layer, net::buffer(m_sspi_impl.shutdown.data()), net::transfer_exactly(m_sspi_impl.shutdown.data().size()), std::move(self));
+          self.complete(m_sspi_impl.handshake.last_error());
+          return;
+      }
+
+      if (m_sspi_impl.shutdown() == detail::sspi_shutdown::state::error) {
+        if (!is_continuation()) {
+          BOOST_ASIO_CORO_YIELD {
+            auto e = self.get_executor();
+            net::post(e, [self = std::move(self), ec, length]() mutable { self(ec, length); });
```
djarek, line 235: Note: posting like this "hides" the associated allocator (it's still correct, but possibly not as performant).

include/boost/windows_sspi/stream.hpp
```diff
@@ -319,6 +365,24 @@ class stream : public stream_base {
...
+      case detail::sspi_shutdown::state::error:
+        ec = m_sspi_impl.shutdown.last_error();
+    }
+  }
+
+  template <typename CompletionToken>
+  auto async_shutdown(CompletionToken&& token) ->
+    typename net::async_result<typename std::decay<CompletionToken>::type,
+                               void(boost::system::error_code)>::return_type {
+    return boost::asio::async_compose<CompletionToken, void(boost::system::error_code)>(
```
djarek, line 382: Shouldn't this be `net::async_compose`?

include/boost/windows_sspi/detail/sspi_impl.hpp
```diff
@@ -371,12 +378,93 @@ class sspi_decrypt {
...
+                                                                     &out_flags,
+                                                                     nullptr);
+    if (m_last_error != SEC_E_OK) {
+      return state::error;
+    }
+
+    // TODO: Avoid this copy
+    m_output_data = std::vector<char>{reinterpret_cast<const char*>(OutBuffers[0].pvBuffer),
+      reinterpret_cast<const char*>(OutBuffers[0].pvBuffer) + OutBuffers[0].cbBuffer};
+    detail::sspi_functions::FreeContextBuffer(OutBuffers[0].pvBuffer);
```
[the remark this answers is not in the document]
laudrup: Thanks. I'm very much aware of this. What is even worse IMO is that minor changes could lead to leaks in much more likely cases, so this should definitely be dealt with.

Having something like a `FreeWhatever()` function in a C library very much calls for being wrapped in a RAII class. The thing is, that I most likely want to wrap these weird `SecBuffer` structs the SSPI library uses in one or more generic classes, so they can be reused in the various functions as well as taking care of cleanup in destructors when needed.

I don't really know exactly how these wrapper classes should to look, so I've decided to leave it as it is for now until I've finished writing all the code where I make use of them. Hope that makes some kind of sense.

include/boost/windows_sspi/detail/sspi_impl.hpp
```diff
@@ -371,12 +378,93 @@ class sspi_decrypt {
...
+    return state::data_available;
+  }
+
+  boost::system::error_code last_error() const {
+    return error::make_error_code(m_last_error);
+  }
+
+  std::vector<char> data() const {
```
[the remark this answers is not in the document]
laudrup: The copy could and should be avoided, but as with the potential memory leak above, I prefer to keep it as it is for now.

It is getting pretty clear, I think, that the classes in `sspi_impl` are sharing some common functionality/traits/stuff and I hope to somehow refactor that so they can share some functionality and avoid duplication.

Not really sure how to handle that, but I hope it will become clearer once I have implemented the missing functionality.

Again, I hope that makes sense, please feel free to tell me if you think this might be the wrong approach.

include/boost/windows_sspi/detail/sspi_impl.hpp
laudrup: I think I only need to change some options in the call to `AcquireCredentialsHandle()` currently residing in the context class to support client certificates, but there is a flag called `ISC_REQ_USE_SUPPLIED_CREDS` which is documented to indicate that:

> Schannel must not attempt to supply credentials for the client automatically.

So it could indeed be relevant and I'll keep that in mind. The documentation is a bit vague to say the least, but I've tried to ensure that it should be fairly easy to write some tests for that once I start looking into certificate handling.

Thanks for pointing this out.

include/boost/windows_sspi/stream.hpp
laudrup: I wrote this with the kind help of @madmongo1 and I must admit I'm not 100% clear on exactly what is going on here.

I might need to either the use the functionality in `boost::beast` or reimplement it here. This is also duplicated a few places which would be nice to avoid if possible.

include/boost/windows_sspi/stream.hpp
laudrup: I'm fairly certain `async_compose` is only available in `boost::asio` whereas the `net` namespace alias is used for things that are available in the networking TS.

This was written with the help of @vinniefalco who specifically pointed this out.

laudrup:
@djarek Thanks a lot for your comments.

I hope it makes sense that I postpone looking into any of them for now, that doesn't mean that I don't appreciate your effort.

PR #11, opened 2020-10-19
Validate server certs

laudrup:


test/handshake_test.cpp
```diff
@@ -27,6 +27,85 @@ TEST_CASE("handshake") {
...
+                                    io_context.stop();
+                                  });
+
+    boost::system::error_code server_error{};
+    server_stream.async_handshake(boost::asio::ssl::stream_base::server,
+                                  [&server_error](const boost::system::error_code& ec) {
+                                    server_error = ec;
+                                  });
+    io_context.run();
+    CHECK_FALSE(client_error);
```
djarek, line 56: You might consider initializing the error_code to an error value, because an empty implementation of `async_handshake` would pass this test :).

laudrup: :-) Good point. Will do.

PR #12, opened 2020-10-29
Server handshake

laudrup:


include/boost/windows_sspi/context.hpp
```diff
@@ -29,41 +31,34 @@ class sspi_handshake;
...
     , m_verify_mode(verify_none) {
   }
 
-  native_handle_type native_handle() const {
-    return m_impl->handle();
-  }
-
   void add_certificate_authority(const net::const_buffer& ca, boost::system::error_code& ec) {
-    m_impl->add_certificate_authority(ca, ec);
+    try {
```
vinniefalco, line 40: `const_buffer` is usually passed by value

include/boost/windows_sspi/detail/context_impl.hpp
```diff
@@ -11,150 +11,43 @@
...
-  }
-
-  HCERTCHAINENGINE ptr = nullptr;
-};
-
 struct context_impl {
   context_impl()
     : m_cert_store(CertOpenStore(CERT_STORE_PROV_MEMORY, 0, 0, 0, nullptr)) {
     if (m_cert_store == nullptr) {
       throw boost::system::system_error(error::make_error_code(boost::winapi::GetLastError()), "CertOpenStore");
```
vinniefalco, line 31: use `BOOST_THROW_EXCEPTION` instead of raw `throw`

include/boost/windows_sspi/detail/context_impl.hpp
```diff
@@ -11,150 +11,43 @@
...
-
-    std::vector<char> buffer(static_cast<std::size_t>(size.QuadPart));
-    boost::winapi::DWORD_ read;
-    if(!boost::winapi::ReadFile(handle.get(), buffer.data(), static_cast<boost::winapi::DWORD_>(buffer.size()), &read, nullptr)) {
-      ec.assign(boost::winapi::GetLastError(), boost::system::system_category());
-      return;
-    }
-
-    add_certificate_authority(net::buffer(buffer), ec);
+  void load_verify_file(const std::string& filename) {
```
vinniefalco, line 49: Why `string` instead of `string_view`?

include/boost/windows_sspi/detail/win32_file.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+  LARGE_INTEGER_ size;
+  if (!GetFileSizeEx(handle.get(), &size)) {
+    throw system_error(GetLastError(), system_category());
+  }
+
+  std::vector<char> buffer(static_cast<std::size_t>(size.QuadPart));
+  DWORD_ read;
+  if (!ReadFile(handle.get(), buffer.data(), static_cast<DWORD_>(buffer.size()), &read, nullptr)) {
+    throw system_error(GetLastError(), system_category());
```
vinniefalco, line 52: What's with all the exceptions? Shouldn't these be `error_code` based?

include/boost/windows_sspi/context.hpp
laudrup: It isn't in the existing `boost::asio::ssl` code and I'm trying to use the existing interface as much as possible:

https://www.boost.org/doc/libs/1_74_0/doc/html/boost_asio/reference/ssl__context/add_certificate_authority.html

include/boost/windows_sspi/detail/context_impl.hpp
laudrup: OK. Will do.

include/boost/windows_sspi/detail/context_impl.hpp
laudrup: Again, this is the interface in `boost::asio::ssl` so I'm trying to use the same.

include/boost/windows_sspi/context.hpp
vinniefalco: Well, the asio code is wrong :)

include/boost/windows_sspi/detail/context_impl.hpp
vinniefalco: asio uses `std::string` because it supports C++03 and the SSL code is ancient. Since you require C++11 (I hope) and Boost, you can use `boost::string_view` instead, which is a better choice.

include/boost/windows_sspi/detail/win32_file.hpp
laudrup: I've decided to use exceptions in the implementation details. IMO it makes the code a bit simpler. I just have to make sure that I catch them and set an error code instead if that's the function the user calls instead of the no-error-code overloads.

Please let me know if there's some reason I shouldn't be doing that. I'm fairly certain I don't have to support cases where exceptions are disabled (and I'm not even sure if that's possible on Windows).

include/boost/windows_sspi/context.hpp
laudrup: It most likely is. I actually think the `boost::asio::ssl` interface should be cleaned up quite a bit, but this is one of the very minor things :-)

include/boost/windows_sspi/detail/context_impl.hpp
laudrup: I've sort of decided to try and make this a "drop in replacement" for `boost::asio::ssl`, so I prefer to just copy the existing interface.

My goal is to be able to take existing `boost::asio::ssl` code and only change a namespace and everything should work as before, so I prefer not to change this.

Of course the caller could most likely call the function the same way whether it's a `std::string` or a `boost::string_view`, but I think it's preferable that the functions look the same.

madmongo1: My 2c here is that if the underlying C API is going to require a null-terminated string and the parameter type to the wrapper is a `string_view`, then the wrapper will need to create a `std::string` (or similar copy) of the text being passed. In this case, having `std::string` as the interface type is no worse, and often better than using a `string_view` - at least in terms of byte copies and possible memory allocation.

If the underlying API is either: 
a) going to make a copy anyway, as in the case of an async initiating function, or
b) or act on the sequence of bytes and return immediately without needing a null termination,

then a `string_view` can avoid the creation of an un-necessary string at the client site.

Having said all this, **as of c++20**, `std::string` becomes `constexpr` if used immediately and discarded (at least according to the standard) so the creation of the `std::string` at the client site can actually be optimised away.

laudrup: @madmongo1 Thanks a lot for your input.

Since the argument is a path to a file in the filesystem, I think the correct type to use would actually be `boost::filesystem::path` or `std::filesystem::path`. That should also handle unicode characters in the filename, but that would of course add a dependency on either `boost::filesystem` or C++17.

As I mentioned, I originally wanted this to be a "drop in replacement" for `asio::ssl`, so I should imply use the same signature as the existing function from `asio::ssl`.

Thinking a bit more about that and after a very valuable talk with @vinniefalco on the Slack channel, I've come to the conclusion that it doesn't make much sense to aim for API compatibility with `asio::ssl` anyway, so I plan to restructure "my" API a bit anyway and will keep this in mind.

Interesting point about C++20 and `constexpr string`. Just as I was learning to not just pass by const reference, but thinking about move, views etc. I have to consider whether a const reference is actually the right thing to do just like it was pre-C++11. C++ is really making things simpler by making them more complicated :-)

PR #28, opened 2021-06-20
Asio : Do not really merge

creativedisassembly:
Still PoC.

Minimal project with lots of hacky things (like including from the result of macro)

Still works on my machine :-)

New MSVC project (I'll figure Cmake out later) with a synchronous ASIO-header only example of a HTTPS Get generator.

If you want to compile locally (and that would be lovely - please give feedback)

1) I'm using 
 - Microsoft Visual Studio Community 2019
- Version 16.10.1
2) Download asio-1-18-2
3) Point the MSVC-Include directories at the Header-only library
4) I only have tried x64 (Debug/Release) so far

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,10 @@
+﻿################################################################################
+# This .gitignore file was automatically created by Microsoft(R) Visual Studio.
+################################################################################
+
+/.vs
+/build/.vs
+/examples/asio-only/.vs/AsioOnly/v16
+/examples/asio-only/Debug
+/examples/asio-only/x64
+/build
--- a/examples/asio-only/AsioOnly.sln
+++ b/examples/asio-only/AsioOnly.sln
@@ -1,0 +1,31 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.31402.337
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "AsioOnly", "AsioOnly.vcxproj", "{A960EC43-B31C-423C-9EE9-9F772A0F4616}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
+		Release|x64 = Release|x64
+		Release|x86 = Release|x86
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{A960EC43-B31C-423C-9EE9-9F772A0F4616}.Debug|x64.ActiveCfg = Debug|x64
+		{A960EC43-B31C-423C-9EE9-9F772A0F4616}.Debug|x64.Build.0 = Debug|x64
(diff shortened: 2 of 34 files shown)


laudrup:
First of all, thanks a lot for your interest in this library. It's really encouraging to not only have some users but even better when someone is willing to make improvements.

Making this library work with standalone asio hasn't really been very high on my list of priorities, but anything that could get more users is of course very welcome.

Considering the actual pull request, I haven't looked at it in detail but it is indeed quite a bit from being something I'm willing to merge. You admit so yourself :-)

I have looked a bit for other libraries that can use both `boost::asio` and standalone asio, but haven't really found anything. Maybe you could find something for inspiration on ways to do it.

Some general points:

- Don't bother too much on the server part if you can avoid it. I plan to rewrite it quite a bit at some point. My initial idea was that this should be a drop in replacement for `asio::ssl` but that didn't turn out to be feasible but it still shows from the clunky API in the server implementation.
- `boost::system` is being used for error codes/exceptions. I assume that could be replaced with `std::system_error` possibly as a compile time configuration option (e.g. `using error_code = std::system_error::error_code` or similar).
- The `boost::winapi` parts can be removed and replaced with our own headers. I would really like to avoid having to include any Windows headers at all at some point.
- I would prefer if the configuration was placed in `config.hpp`. Hopefully most can be done by some `using` aliases as I would really like to avoid too many macros i possible.
- Ideally it should be possible to choose whether to use standalone asio at CMake configure time passing something like `-DASIO_STANDALONE` to CMake defaulting to using `boost::asio` like it is now.

Thanks for the effort so far and good luck with the code.

creativedisassembly:
Some general answers :-)

- Don't bother too much on the server part if you can avoid it. I plan to rewrite it quite a bit at some point. My initial idea was that this should be a drop in replacement for asio::ssl but that didn't turn out to be feasible but it still shows from the clunky API in the server implementation.
**It's still very nice. Look how easy an HTTPS request can be..(Assuming you receive the reply in one go)**
**
	std::string reply;
	TemplateDefinitions map{ {"host", host}, {"path", path} };
	const auto rqst = ApplyTemplate(BasicGetRequest, map);
	std::cout << rqst << std::endl;
	try {
		stream << rqst;
		stream >> reply;
	}
	catch (const net::error_code& ec)
	{
		std::cerr << ec.message() << std::endl;
	}
**

- I assume that could be replaced with std::system_error possibly as a compile time configuration option (e.g. using error_code = std::system_error::error_code or similar).
**Really there are only two uses for errors/exception in the code. Simple error_codes, so asio::error_code slots in, and an error with an additional string. Easy to emulate**

- The boost::winapi parts can be removed and replaced with our own headers. I would really like to avoid having to include any Windows headers at all at some point.
**Its already like that, I created a new sub-directory called winapi. The includes pick this up instead of boost::winAPI.**
- I would prefer if the configuration was placed in config.hpp. Hopefully most can be done by some using aliases as I would really like to avoid too many macros i possible.
**Will do**
- Ideally it should be possible to choose whether to use standalone asio at CMake configure time passing something like -DASIO_STANDALONE to CMake defaulting to using boost::asio like it is now
**Already done**

The next thing to do is fix up the boost build again. I sort of ploughed-through to get the examples compiling, and the boost-version has been left broken, just with macros, where it needs fixing.

It has to be said, that part of me wonders if we should drop the boost-bit, and just call it an asio-fork. However, I want your fixes (always the trade-off), so I feel like it's worth the try even if it is macro-hell :-)

PR #31, opened 2021-06-24
Avoid encrypt buffer copy

laudrup:


madmongo1 (approved):
Looks reasonable, as long as the sspi_buffers stays alive during the composed operation.

laudrup:
> Looks reasonable, as long as the sspi_buffers stays alive during the composed operation.

The `sspi_buffers` are kept inside the `sspi_encrypt` class which is alive during the entire lifetime of the `stream` class, so that shouldn't be a problem.

Part of this change was trying to avoid having to worry about these issues in the composed operation code which also makes it a lot easier to avoid unnecessary copies.

I'm quite happy with how the refactoring turned out to make the code a lot more readable IMO. Thanks for having a look before I start to use the same ideas for the decrypt, handshake and shutdown operations.

PR #36, opened 2021-07-01
Cleanup shutdown implementation

laudrup:


include/boost/wintls/detail/async_shutdown_impl.hpp
```diff
@@ -34,23 +34,23 @@ struct async_shutdown_impl : boost::asio::coroutine {
       return entry_count_ > 1;
     };
 
+    net::const_buffer buffer = sspi_impl_.shutdown(ec);
```
vinniefalco, line 37: Where is the storage for the contents of `buffer`?

laudrup: That's embarrassing. I was too satisfied with my "smart pointer like wrapper" for pointers returned from the SSPI function that I forgot to use it correctly.

Will fix, thanks a lot for spotting this.

PR #47, opened 2021-08-05
Add a test doing handshake on a moved stream

laudrup:
In order for the pattern used by the asio server examples to work, a
stream must be movable to add a test to demonstrate that this
currently *doesn't* work.

### DIFF
--- a/test/stream_test.cpp
+++ b/test/stream_test.cpp
@@ -33,6 +33,55 @@ public:
   }
 };
 
+TEST_CASE("moved stream") {
+  boost::asio::io_context ioc;
+
+  wintls_server_context server_ctx;
+
+  boost::wintls::stream<test_stream> server_stream(ioc, server_ctx);
+  //
+  // Replacing the above with this causes the test to hang with the
+  // serving waiting for the client to send it data which will never
+  // happen as the client has already sent a complete TLS CLIENT HELLO
+  //
+  // boost::wintls::stream<test_stream> stream(ioc, server_ctx);
+  // boost::wintls::stream<test_stream> server_stream(std::move(stream));
+
+  wintls_server_context client_ctx;
+
+  boost::wintls::stream<test_stream> client_stream(ioc, client_ctx);
+  //
+  // And replacing the above with this causes the client to hang for
+  // the same reason
+  //
+  // boost::wintls::stream<test_stream> stream(ioc, client_ctx);
+  // boost::wintls::stream<test_stream> client_stream(std::move(stream));
+
+  client_stream.next_layer().connect(server_stream.next_layer());
+
+  boost::system::error_code client_ec{};
+  boost::system::error_code server_ec{};
+
+  server_stream.async_handshake(boost::wintls::handshake_type::server,
+                                [&server_ec, &server_stream](const boost::system::error_code& ec) {
+                                  server_ec = ec;
+                                  if (ec) {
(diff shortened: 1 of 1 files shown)


laudrup:
This is WIP and not intended to be merged.

This contains a commit with a test that demonstrates how the `wintls::stream` will hang after having been moved.

The reason the test hangs is, that the underlying call to to `InitializeSecurityContext` will return `SEC_I_CONTINUE_NEEDED` which means the function claims it hasn't received all the data required for a complete TLS CLIENT HELLO message.

As can be seen in the test, this is both the case when using the stream to do a server handshake as well as a client handshake with similar symptoms.

The weird thing is that the function has received all the data needed. I have looked into what is being passed to the function when it's working correctly (i.e. not being moved) and after it's been moved and it does receive the same values (ignore pointer differences and TLS message differences because of randomness being exchanged).

The exact same thing happens if using a TCP socket with an example program. Doing so I have sniffed the traffic with Wireshark and the server does receive a complete TLS CLIENT HELLO. The payload seen in wireshark is exactly the same as the one being passed to the function.

I sort of suspect the move might invoke some undefined behavior I cannot spot. The weird thing is that the symptoms are the same with both MSVC and Clang both in debug and release builds. I have also tried using address sanitizer just to see if that might complain about something.

It might also be another Windows API weirdness, but although the Windows API is weird to put it mildly, I still have no idea what vould be causing this.

madmongo1:
> In order for the pattern used by the asio server examples to work, a stream must be movable

At face value this is false.

Unless you mean some other kind of non-asio stream?

madmongo1:
Whether the code should work or not will depend on whether your  `boost::wintls::stream` is moveable, which will in part depend on whether the test_stream us moveable.

Does  `boost::wintls::stream` perform any initialisation in its constructor, perhaps offering an address to the windows API?

laudrup:
@madmongo1 Thanks. I was thinking about the `boost::asio::ssl` server example. The async example depends on the `asio::ssl::stream` being movable.

I'm sure the example could be written some way to avoid depending on that, but I'd really prefer to figure out a way to make stream movable no matter what.

laudrup:
@madmongo1 The stream doesn't use any Windows API addresses directly as far as I can tell.

It does create two (unique) pointers to some handles which are then used after the stream has been moved, but I cannot see why that would be a problem?

Also, the stream does have some member classes that has even more member classes of their own, but I haven't been able to spot anything that might be problematic with any of them.

Again, I hope I'm overlooking something. Thanks a lot.

vinniefalco:
Movable I/O objects are preferable as it allows you to transfer ownership of the underlying socket to a newly constructed "connection" object. For example, transferring from a generic HTTP connection object to a specific WebSocket connection object.

laudrup:
@vinniefalco Indeed. I think it would be most unfortunate if this object couldn't be moved which is also why it annoyed me that I couldn't figure out what might cause that

vinniefalco:
Default move construction will not work for this field, as it stores a pointer:
https://github.com/laudrup/boost-wintls/blob/041f8b0f876c78768a9e81454838e392f0e35710/include/boost/wintls/detail/sspi_handshake.hpp#L42
The newly constructed stream will be pointing to a member of the moved-from stream, with predictable results.

laudrup:
Thanks for the very useful feedback. This has now been fixed on the master branch.

PR #55, opened 2022-06-28
Feature: Client SSL Validation

### LINKED ISSUE #53: [Feature Request] Verify client ssl in server
In asio openssl implementation, when writing a server, one can use
```cpp
context_->set_verify_mode(boost::asio::ssl::verify_peer | boost::asio::ssl::verify_fail_if_no_peer_cert);
```
Or use callback to verify client has the right certificate present, and rejects all clients that do not present the right cert.
Is there a way to do the same in wintls? If not how to support it?

youyuanwu:
Enables server to request and validate client certificate.

First batch for #53.

### DIFF
--- a/include/boost/wintls/detail/sspi_handshake.hpp
+++ b/include/boost/wintls/detail/sspi_handshake.hpp
@@ -28,11 +28,15 @@ namespace detail {
 
 class sspi_handshake {
 public:
+  // TODO: enhancement: done_with_data and error_with_data can be removed if 
+  // we move the manual validate logic out of the handshake loop.
   enum class state {
-    data_needed,
-    data_available,
-    done,
-    error
+    data_needed,      // data needs to be read from peer
+    data_available,   // data needs to be write to peer
+    done_with_data,   // handshake success, but there is leftover data to be written to peer
+    error_with_data,  // handshake error, but there is leftover data to be written to peer
+    done,             // handshake success
+    error             // handshake error
   };
 
   sspi_handshake(context& context, ctxt_handle& ctxt_handle, cred_handle& cred_handle)
@@ -50,7 +54,7 @@ public:
     SCHANNEL_CRED creds{};
     creds.dwVersion = SCHANNEL_CRED_VERSION;
     creds.grbitEnabledProtocols = static_cast<DWORD>(context_.method_);
-    creds.dwFlags = SCH_CRED_MANUAL_CRED_VALIDATION;
+    creds.dwFlags = SCH_CRED_MANUAL_CRED_VALIDATION | SCH_CRED_NO_DEFAULT_CREDS;
 
     auto usage = [this]() {
       switch (handshake_type_) {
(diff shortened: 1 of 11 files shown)


laudrup:
Thanks a lot for this.

I haven't looked at it in detail but as far as I can tell it looks really promising. Well done.

Please do continue the good work and let's have a discussion about this when you feel like it's something that could be merged.

youyuanwu:
I wrote a plan for implement this feature in a few steps (PRs) to separate the concerns in the linked issue.
This PR is ready to be checked in as is, improvements will be in subsequent PRs. I noticed there is a clang build or checker error I will fix asap.

include/boost/wintls/detail/sspi_handshake.hpp
```diff
@@ -176,31 +192,37 @@ class sspi_handshake {
...
           if (last_error_ != SEC_E_OK) {
             return state::error;
           }
-
-          cert_context_ptr remote_cert{ctx_ptr, &CertFreeCertificateContext};
-
-          last_error_ = static_cast<SECURITY_STATUS>(context_.verify_certificate(remote_cert.get()));
-          if (last_error_ != SEC_E_OK) {
-            return state::error;
+        }else{
```
laudrup, line 211: Very minor thing but please add spaces between the braces:. `} else {`

include/boost/wintls/detail/sspi_handshake.hpp
```diff
@@ -242,6 +264,26 @@ class sspi_handshake {
   }
 
 private:
+  SECURITY_STATUS manual_auth(){
+    if(!context_.verify_server_certificate_)
+    {
```
laudrup, line 269: Also a very minor thing but please move the brace up to the line with the `if`.

test/handshake_test.cpp
```diff
@@ -129,6 +133,194 @@ TEST_CASE("certificates") {
...
+
+    auto server_error = errc::make_error_code(errc::not_supported);
+    server_stream.async_handshake(asio_ssl::stream_base::server,
+                                  [&server_error](const boost::system::error_code& ec) {
+                                    server_error = ec;
+                                  });
+    io_context.run();
+    // client handshake is failed by server
+    CHECK(client_error);
+    // Note: error code mapping is not working, so compare string.
```
laudrup, line 165: What does it mean that "error code mapping is not working"? Could it be made to work?

youyuanwu: I added more info in the new version. It seems like the top bits of the error code is used for something else. I will need to debug into the openssl code further to find out why, but it is out of scope of this PR.

youyuanwu:
> There are a few TODOs in the code. I assume these are the ones you plan to resolve in later PRs?
> 
Yes. This PR is not breaking any existing APIs and functionalities, but some TODOs might. So better make them in separate PRs. 
> The only thing that seems to be missing is documentation but I assume that will be added once server and client certs have been moved to separate functions?
>
Yes. The verfiy_server_certificate API will be changed to use verfify_mode instead so it is better to document in that future PR.

laudrup:
@youyuanwu 

Since this isn't breaking anything (hopefully) I guess this could be merged into the master branch as it is if you prefer?

If so, I think this should be squashed to a single commit and I would really prefer if you could use a more descriptive commit message. Please write a few lines on what this does and my importantly why this is needed.

Not a long novel, just a bit more than a single line. If you look at the history you can hopefully get an idea of what I mean. Not that my commit messages are perfect though :-)

youyuanwu:
Squashed as a single commit. But I think for if you merge a PR on github, you can choose "merge squash" mode, and the PR title and description will be set as the commit tile and content.
Please merge this so that I can work on top of this. Thanks.

Issue #56, opened 2022-06-29
Various Remarks


jens-diewald:
Hi!
First of all, thanks for putting together this very useful library.
I have been using beast+asio(and hence openssl)+[djarek/certify](https://github.com/djarek/certify) for a simple https client so far.
But i do prefer your approach and would like to replace openssl+certify by your library.
Here are a few notes from first tests:

1. Server hostnames are not verified during certificate validation. (Testcase: wrong.host.badssl.com:443)
 This should be a relatively easy fix if i am not missing something. I can add a small pull request later for further details.
2. There are several more (probably less important) test cases on badssl.com, that are not handled as the site suggests they should be.
 Checking a revocation list should be relatively straight forward, not sure about the other cases.
3. Trying to connect to client.badssl.com:443 yields a failed assert due to client certificates not being implemented. I think it would be nicer to throw an exception in such a case, so the caller can recover from the error. (Since there is #55 now, this particular case is probably not a problem for long though :) )
4. I needed to link against secur32.lib and crypt32.lib to use the library (You have listed even two more in your cmake config). This should be documented somewhere. Maybe we can even have some pragma comment lib for convenience?
5. I am not sure how far you are aiming for compability with asio/ssl, but i ran into a few differences. This is no big issue for me, but a higher compatibility would be nice to allow simple cross platform code to use wintls on windows and asio with openssl on unix systems.
   - wintls has context::use_default_certificates(bool) instead of context::set_default_verify_paths(). I think use_default_certificates is the better (more general) name and i suppose windows does not really have "verify paths". But maybe the other function could be introduced as an alias for compability?
   - wintls has context::verify_server_certificate(bool) instead of context::set_verify_mode(verify_mode). Here the latter is more general, but as i understand it wintls does not (yet) have all the functionality that asio provides here. It would be nice to somewhat align this with asio in the future. (As also mentioned here: https://github.com/laudrup/boost-wintls/issues/53#issuecomment-1169608161)
   - wintls uses some scoped enums (wintls::method, wintls::handshake_type) which are different from the asio equivalents. I generally like the wintls way much better, not sure what could be done here...
   - There is no wintls::error::stream_truncated. The asio equivalent is a common error that i have to handle. (Not sure what wintls yields in these cases yet.)
[...]

PR #57, opened 2022-06-29
Pass Hostname to Certificate Verification

jens-diewald:
Introduce verification of host names as mentioned in 1) in #56 

I suppose using the hostname set in set_server_hostname for this is fine? The [documentation](https://wintls.dev/classes.html#_CPPv4N5boost6wintls6stream19set_server_hostnameERKNSt6stringE) already states that this is also used for certificate verification.

The necessary conversion to wstring is unfortunate, but i do not see a better way to do this..?

Ideally there would be a test case for this and possibly more cases in certificate_test.cpp, i guess.
Just taking certificates from badssl.com would be a simple approach. But they will expire at some point and would have to be renewed which depends on the continued avilability of that site.

### DIFF
--- a/test/certificate_test.cpp
+++ b/test/certificate_test.cpp
@@ -10,6 +10,7 @@
 
 #include <boost/wintls/certificate.hpp>
 #include <boost/wintls/error.hpp>
+#include <boost/wintls/detail/context_certificates.hpp>
 
 #include <fstream>
 #include <iterator>
@@ -71,4 +72,16 @@ TEST_CASE("import private key") {
 
   boost::wintls::delete_private_key(name, ec);
   CHECK(ec.value() == NTE_BAD_KEYSET);
+}
+
+TEST_CASE("verify certificate host name") {
+  const auto cert = boost::wintls::x509_to_cert_context(net::buffer(test_certificate), boost::wintls::file_format::pem);
+  boost::wintls::detail::context_certificates ctx_certs;
+  ctx_certs.add_certificate_authority(cert.get());
+  // success case: host name is not verified when parameter is empty string
+  CHECK(ctx_certs.verify_certificate(cert.get(), "") == ERROR_SUCCESS);
+  // success case: test_certificate contains the host name "localhost"
+  CHECK(ctx_certs.verify_certificate(cert.get(), "localhost") == ERROR_SUCCESS);
+  // fail case: incorrect host name 
+  CHECK(ctx_certs.verify_certificate(cert.get(), "wrong.host") == CERT_E_CN_NO_MATCH);
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,9 @@
 __pycache__
-build
+/out/build
+
+# Visual Studio Code generated files
 .vscode
+
+# Visual Studio generated files
+/.vs
+/CMakeSettings.json
--- a/include/boost/wintls/certificate.hpp
+++ b/include/boost/wintls/certificate.hpp
@@ -218,7 +218,7 @@ inline void delete_private_key(const std::string& name, boost::system::error_cod
  * existing or non accessible key will result in unexpected behavior
  * when used with a @ref stream operating as a server.
  *
- * @param cert The certifcate to associate with the private key.
+ * @param cert The certificate to associate with the private key.
  *
  * @param name The name of the private key in the default cryptographic key provider.
  *
--- a/include/boost/wintls/context.hpp
+++ b/include/boost/wintls/context.hpp
@@ -13,6 +13,8 @@
 #include <boost/wintls/detail/config.hpp>
 #include <boost/wintls/detail/context_certificates.hpp>
 
+#include <string>
+
 namespace boost {
 namespace wintls {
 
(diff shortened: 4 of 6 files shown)


laudrup:
Thanks a lot for this.

I would really appreciate it if you could add a test for this. Both for failing and successful hostname validation.

I hope it should be possible to extend the existing tests.
Thanks.

jens-diewald:
I have added a simple test case. I am not super happy with it, as it really only checks host name verification. I think it would be nice to have a more general facility to test certificate verification, but i am not sure how that could be done right now and it would certainly be more work.

Also i have adjusted my original commit to your indentation and brace wrapping. I did not realize this before. Having a .clang-format file in the repository would help a lot to avoid such mistakes. Could you maybe provide one, or do you happen to use some other form of formatting rules?

By the way: The Unittests in error_test.cpp that are checking error messages as strings fail for me, as the strings are in german for me.

test/certificate_test.cpp
```diff
@@ -72,3 +73,16 @@ TEST_CASE("import private key") {
   boost::wintls::delete_private_key(name, ec);
   CHECK(ec.value() == NTE_BAD_KEYSET);
 }
+
+TEST_CASE("verify certificate host name")
+{
```
laudrup, line 78: Really minor thing but please move the starting brace to the previous line

laudrup:
> I have added a simple test case. I am not super happy with it, as it really only checks host name verification. I think it would be nice to have a more general facility to test certificate verification, but i am not sure how that could be done right now and it would certainly be more work.
> 

Thanks a lot. It's definitely better than nothing.

A more general way of testing certificate verification would definitely be a good idea but I agree that it is out of scope for this change. You are of course very welcome to try and implement it ;-)

> Also i have adjusted my original commit to your indentation and brace wrapping. I did not realize this before. Having a .clang-format file in the repository would help a lot to avoid such mistakes. Could you maybe provide one, or do you happen to use some other form of formatting rules?
> 

I actually have a clang-format I just haven't commited it for some reason. I have a bit mixed opinions about whether it's a good idea. Auto formatting code is mostly fine but really annoying in the cases where it doesn't work.

I'll try to see if I can add what I have. I probably won't do anything to enforce it's usage though.

> By the way: The Unittests in error_test.cpp that are checking error messages as strings fail for me, as the strings are in german for me.

That's pretty bad. I hadn't considered that and tests definitely shouldn't depend on the locale of the OS.

I guess I could either see if I could find a way of testing that the errors actually do end up as `SECURITY_STATUS` errors which is really the point of the tests or do some ugly changing of locale for the test process.

If you have any idea on how to solve this please do let me know. In any case it would be nice if you could create an issue for this so I don't forget about it.

Anyway, I think I can merge this if you feel confident it should be OK?

jens-diewald:
> I actually have a clang-format I just haven't commited it for some reason. I have a bit mixed opinions about whether it's a good idea. Auto formatting code is mostly fine but really annoying in the cases where it doesn't work.
> 
> I'll try to see if I can add what I have. I probably won't do anything to enforce it's usage though.

That would be great! I had to actively work against my auto formatting and still did miss the one brace you mentioned above ;)

> > By the way: The Unittests in error_test.cpp that are checking error messages as strings fail for me, as the strings are in german for me.
> 
> That's pretty bad. I hadn't considered that and tests definitely shouldn't depend on the locale of the OS.
> 
> I guess I could either see if I could find a way of testing that the errors actually do end up as `SECURITY_STATUS` errors which is really the point of the tests or do some ugly changing of locale for the test process.
> 
> If you have any idea on how to solve this please do let me know. In any case it would be nice if you could create an issue for this so I don't forget about it.

I will create an issue. Off the top of my head i have no good idea either...

> Anyway, I think I can merge this if you feel confident it should be OK?

Yes, should be okay :+1:

laudrup:
@jens-diewald 

It all looks very good. One final thing.

In order to merge this I think it should be squashed into a single commit. It is one change after all.

It seems like I had disabled squashing and merging in github but I have enabled that now. Don't know if that affects you.

Could you please squash the commits into a single commit and provide a bit more descriptive commit message?

A header like "Pass Hostname to Certificate Verification" is fine, but please add a few lines on why this change was necessary. I like descriptive commit messages.

Thanks once again.

jens-diewald:
I have rebased my Branch to the current master, squashed the commits and added a descriptive message.

However, i also started thinking about the host name encoding.
Do you expect wintls to support non-Ascii encodings (i.e. IDNs)?
My usage of `std::mbstowcs` converts the host name based on the current locale. [According to the documentation](https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/ns-wincrypt-httpspolicycallbackdata) it would be fine to pass non-Ascii strings to `CertVerifyCertificateChainPolicy` and they would internally be converted to punycode and handled reasonably.
For setting the SNI host name via `InitializeSecurityContextA`, i expect that always ANSI (CP1252) encoding is assumed. But i can not find any documentation if that function would actuall handle non-Ascii strings... The corresponding parameter `pszTargetName` is not even described [here](https://docs.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-initializesecuritycontexta) and there is no relevant mention for [the W variant](https://docs.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-initializesecuritycontextw).
Afert all, i am not even sure, `asio::ip::tcp::resolver` supports non-Ascii host names? It does not seem to work for me, even though it seems it seems to be implemented in terms of [getaddrinfo](https://docs.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-getaddrinfo) for me, which should support ANSI host names.
The Asio documentation does not say anything regarding encoding. Do you know more about this?

If you think only support Ascii host names should be supported in wintls, i still think this PR is good to merge.
If you expect ANSI encodings to work, i should change the `std::mbstowcs` to something else and add another test case.

jens-diewald:
Ah, i wrongly assumed the Windows A functions always expect CP1252. They do expect the active code page though, so using std::mbtowcs should be fine (in line with directly passing the host name to InitializeSecurityContextA).
(Oh my, why can't everything be unicode :) ).

Still, do you know if asio should be able to handle non-Ascii? It does not seem to work for me.

PR #58, opened 2022-07-01
Update link libraries

laudrup:


include/boost/wintls/detail/config.hpp
```diff
@@ -11,6 +11,11 @@
 #include <boost/config.hpp>
 #include <boost/asio.hpp>
 
+#if defined(BOOST_MSVC)
```
jens-diewald, line 14: Clang supports this as well (actually on [Windows](https://clang.llvm.org/docs/UsersManual.html#microsoft-extensions) and on [Unix](https://clang.llvm.org/docs/LanguageExtensions.html#specifying-linker-options-on-elf-targets)).
Boost.Config has this check: https://github.com/boostorg/config/blob/develop/include/boost/config/auto_link.hpp#L101
I would suggest doing it something like this? Maybe leave out Borland, that shouldn't be relevant, right?

laudrup: Thanks a lot for commenting on my broken commits :-)

I actually thinks it makes better sense to just check for anything other than gcc/mingw.

These are the only compiler I test with and which we have CI jobs for. If anyone wants to add support for something else they are of course more than welcome.

Thanks once again. Your input is much appreciated.

jens-diewald: Ah yes, that is easier, good idea!

PR #64, opened 2022-07-25
Clang format

jens-diewald:
While working on #63, clang-format did not work for me without any helpful error message.
(Besides three Options that are not yet available in Clang 13 which i am using, but i can remove these locally, that is fine.)
I just figured out that removing the line `BasedOnStyle: None` solves the problem for me. Also, according to the clang format documentation the value None does not seem to exist for that option.

After getting it to work, i have also made some more changes to align the settings a bit more with the existing code.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,4 @@
 Language: Cpp
-BasedOnStyle: None
 AlignAfterOpenBracket: Align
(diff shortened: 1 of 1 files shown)


laudrup:
Nice. Thanks a lot.

Issue #65, opened 2022-08-12
(x86 build error) ambiguous overloaded operator[] with handshake_input_buffers/handshake_output_buffers


leo-liu:
Building the wintls example with VS2022 on 32bit x86, it shows that there are some build error on overloaded `operator[]` with type `handshake_input_buffers`/`handshake_output_buffers`. I also tried LLVM-cl with the same error. However, the x64 build is OK.

I inspected the code and found out the ambiguous overloaded functions are:
```c++
template <std::size_t N>
class sspi_buffer_sequence {
public:
  //...

  operator PSecBufferDesc() {
    return &sec_buffer_desc_;
  }

  sspi_buffer& operator[](size_t i) {
    return buffers_[i];
  }

  const sspi_buffer& operator[](size_t i) const {
    return buffers_[i];
  }

  // ...
private:
  SecBufferDesc sec_buffer_desc_;
};
```
Then a call to `input_buffers_[0].pvBuffer` may cause a build error.

It seems that the bug is concealed by the different pointer size or integer size on x64.

leo-liu:
Personally I think it is not a good idea to overload implicit conversion from `sspi_buffer_sequence` to `PSecBufferDesc`. The bug can be fixed by replacing `operator PSecBufferDesc()` to an explicit function call.

laudrup:
Closed with #66

PR #66, opened 2022-08-12
fix sspi_buffer_sequence ambiguous overload on x86

leo-liu:
it should fix issue #65


laudrup:
Thanks for spotting and fixing this.

I agree that implicit conversions are most often not a good idea and I have no idea why I thought it would be a good idea in this case.

I didn't really think 32 bit support was relevant anymore but it's a bug anyway so thanks for the fix and for letting me know that 32 bit is actually supported with this fix.

Much appreciated.

PR #67, opened 2022-08-19
Add OCSP Test

jens-diewald:
Revocation checking was implemented in #63, but the new unit test introduced there only covered CRLs.
OpenSSL includes a small OCSP responder and as the unit tests already depend on openssl, i realized that this would be a relatively simple way to add a test case for that.
It turned out not as simple as i hoped, but it works now.
There are still some peculiarities, however. See my comments in the code for details. Particularly, for some reason it takes about 15 seconds (at least on my machine) until the OCSP responder actually responds. That makes the unit tests run much longer than before. With regard to the CI that runtime does not really matter, but it might be a bit annoying if they are frequently run locally.

Somewhat independent of this, i have cleaned up the generate_certs script a bit while adding the certificates neccessary for the new test.


laudrup:
Hi @jens-diewald 

Thanks a lot for your efforts on this. It's really appreciated.

Regarding the much longer time to run to the test, to be honest, I'm not really happy with that. I prefer that unittests run in a matter of seconds. Otherwise I don't think you can really consider them unittests.

That doesn't mean that the tests are not useful though, but they should probably be considered more as "integration tests" or similar.

A solution to that might be to use Catch2 tags to mark these tests as integration tests. There's even an example of that in the [Catch2 documentation](https://github.com/catchorg/Catch2/blob/v2.x/docs/test-cases-and-sections.md#special-tags). This way they will not be run as part of the default test suite but can still be run if specifying so on the command line. One of us could then update the github actions workflow so it is only run for one of the platforms/versions (probably the latest).

As you have most likely noticed the tests are currently failing. As far as I can tell it seems like there's an issue finding the boost libraries for the 141 and ClangCL toolsets. That has probably not been an issue as long as only the headers were used but since this PR adds a dependency on `boost::filesystem` we also need to link with the compiled libraries.

Not really sure what to do about that. I had really hoped that the static libraries of boost would work in the CI pipeline although I haven't tested it so it's not an issue with your specifically. It ought to be working since it might be needed for other changes later on.

I guess I wouldn't mind dropping the tests for v141 toolset. Not sure how much that matters as I'm not that much experienced with the Windows platform. Maybe you have an opinion?

I would *not* like to drop support for the Clang compiler though so I guess it needs to be fixed somehow but I'm not really sure how at the moment. Maybe an issue with the [install-boost](https://github.com/MarkusJx/install-boost) action being used?

jens-diewald:
Hi, thanks for the feedback.
I am happy to work on this a bit, it is an interesting project and i am still a bit confused that this library did not exist before ;)

Making this test case into an "integration test" seems reasonable, i will look into that.
(Or maybe i can understand why the OCSP responder takes that long to start up and fix that... I don't see a good reason why it does.)

As for the failed tests, it looks like install-boost does determine the msvc toolset version for which the static libs are installed by its parameter platform_version (2016->vc141, 2019->vc142, 2022->vc143), so this should be set accordingly.
Binaries for Clang do not seem to be available via install-boost, but it is possible to link against msvc binaries with clang, so we can just use vc142 for this. We will have to tell cmake/FindBoost to link against these libraries then, but there should be a flag for that.
I have never used github actions, but it looks relatively straight forward. I will try to push a fix for the linking issue later.

jens-diewald:
I did get the test working again through some trail and error.
I had to exclude some combinations of boost and toolset, becauase install-boost does only provide v143 for recent versions (>=1.78) and v141 only for older versions (<1.78).
One could do some more tests by taking advantage of the fact, that the toolset versions are backwards compatible, i.e. link against v142 libs when compiling with v143 for example, but i do not think that this is too important.

I do not have time to do anything further in the next 2-3 weeks, but i will come back to this after that.

laudrup:
Hi @jens-diewald 

Thanks a lot for this and sorry for not getting back to you until now.

I can indeed see you had the "pleasure" of trial-and-error getting the CI pipeline to work :-) It is quite frustrating and good to see you made it work in the end. The change looks good and is quite an improvement in general so thanks for that.

You mentioned that the tests took around 15 minutes to run or so but it doesn't look like that if I look at the jobs. Do you know why that might be?

Thanks once more for you work on this project. It's much appreciated.

jens-diewald:
Hi again!

It is 15 seconds, not 15 minutes, this is also the case for the CI jobs. For example it says `Test #14: check certificate revocation .....   Passed   17.18 sec` in the `Run Jobs` output [here](https://github.com/laudrup/boost-wintls/actions/runs/2913019538/jobs/4640314340).
So maybe it is not as bad as you thought. But i still do not really get why it should take 15 seconds to start up that simple server  and plan to look into that. I would do a separate pull request for that if i figure something out.
So if you are fine with 15 seconds, we can merge this as it is or i can still change the test case to an "integration test".

laudrup:
Hi again :-)

Don't know why I had read that as 15 minutes. Agreed that this is not too bad but I still think these tests should be tagged as integration tests so they can easily be excluded during development where you most often want to build and run the tests frequently.

If you could look into tagging these tests as integration tests I will merge this PR and then I think I should better make a new release soon.

Thanks a lot.

jens-diewald:
Tagging the test as an integration test was easy and works as expected.
I do not know how to pass additional command line arguments through ctest, however.
Therefore i did not manage to re-enable the test in the CI yet. Do you know of a simple way to do that?

laudrup:
> Tagging the test as an integration test was easy and works as expected. I do not know how to pass additional command line arguments through ctest, however. Therefore i did not manage to re-enable the test in the CI yet. Do you know of a simple way to do that?

Thanks once more.

I would think that enabling/disabling the tests in CTest would be supported by using the `catch_discover_tests()` function which is already called in the `CMakeLists.txt` file in the `test` directory. I think that would add some labels that can be used for controlling which tests to run.

I haven't looked into it in detail though but I might do so unless you get to do it first.

laudrup:
@jens-diewald 

That's perfectly fine. I don't use CTest during development myself either so it is probably mostly being used by the CI jobs.

I'll merge this now. Thanks a lot for taking your time to implement this. I really appreciate it.

PR #74, opened 2022-12-14
add shutdown test cases

jens-diewald:
This should not be merged in the current state but is rather supposed to show of a difference between Asio.SSL and WinTLS: Asio will produce a stream_truncated error on shutdown, if the remote did not properly shut down the protocol while WinTLS will not.

Note https://stackoverflow.com/questions/25587403/boost-asio-ssl-async-shutdown-always-finishes-with-an-error

Also, according to that post, Asio should produce an error if it initiated the shutdown and it goes through without problem.
This does not seem to be the case (anymore?).
This is relevant for the https examples in WinTLS. Currently the https async example handles asio::error::eof (as do the beast examples async+sync), while the sync example does not.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -33,6 +33,7 @@ set(test_sources
   sspi_buffer_sequence_test.cpp
   stream_test.cpp
   decrypted_data_buffer_test.cpp
+  shutdown_test.cpp
 )
 
 add_executable(unittest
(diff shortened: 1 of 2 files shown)


laudrup:
Thanks for looking into this.

I'll be happy to merge this once it's done.

jens-diewald:
I close this in favor of #78

PR #76, opened 2022-12-14
Remove functional include and use unique_ptr more efficiently

jens-diewald:
A technicality i came across while looking through the code.
If a unique_ptr type is always used with the same deleter function, it should be part of the deleter type instead of using function pointers or even std::function.
This has the advantage, that we do not have to repeat the deleter function for every instantiation and it is also a little more efficient, as not every unique_ptr has to save the additional function pointer or function object within it. (sizeof(cert_context_ptr) is now 8 and was 16 before.)

### DIFF
--- a/include/boost/wintls/certificate.hpp
+++ b/include/boost/wintls/certificate.hpp
@@ -41,13 +41,19 @@ struct crypt_key {
   HCRYPTKEY ptr = 0;
 };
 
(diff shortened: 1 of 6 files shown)


laudrup:
Sorry, I had completely forgotten about this.

Looks really great. Thanks a lot for your contribution.

Issue #79, opened 2023-11-07
non-boost version support ?


windowsair:
Hi, 
As you know, boost is a large dependency (in my own case), so I'm looking to see if there's a winstl for non-boost, as non-boost Asio has done.

Note that the minimum supported version is C++14, so any boost dependencies on boost-wintls itself can be replaced with STL. 

I have now implemented a non-Boost version of winstl. If you think it's appropriate, I can create a PR and further discuss the specific details :)

laudrup:
Hi @windowsair,

I would definitely be interested in that. Some effort in to doing exactly that [has been tried previously](https://github.com/laudrup/boost-wintls/pull/28) but for some reason it was never finished.

I don't think there's much in the actual code that really depends on other boost libraries than boost.asio. If there is, then that should hopefully be fairly trivial to remove/rewrite and I definitely wouldn't mind that.

If I'm no mistaken one of the main issues is mostly the tests. I make use of the custom `test_stream` class from boost.beast and I would of course still very much like to not only keep the tests but also have them enabled for a non-boost dependent version of the library.

Not sure if it might be worth it to try and implement our own test_stream class or if a non-boost version would simply still depend on boost for the tests. Implementing a test_stream for asio is not trivial but definitely not impossible either.

But first of all feel free to open a PR and let's take it from there.

Thanks a lot.

PR #80, opened 2023-11-13
Add support for standalone Asio

windowsair:
Hi, @laudrup

As we discussed in #79 , the pr is a simple attempt to add support for the non-boost version of wintls. It only contains modifications to the core code, excluding build system and test code. This is still a crude attempt, but should be sufficient for the header only version.

The main tricks are contained in `config.hpp`. Use C++ using alias to distinguish boost and non-boost versions, and use the `WINTLS_USE_STANDALONE_ASIO` macro to control whether to use the boost version.  Some boost functions and macros for assertions have been rewritten, and currently we don't depend on any boost components.

As you say, beast currently depends on boost-asio.  So in my opinion, if we use a library like beast, we can just use the boost version of winstl. Well, I'm not sure if there will be a problem mixing boost-asio with standalone-asio in beast.

There's definitely a lot that can be improved here. Looking forward to your review, thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.15)
 
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
 
+include(FetchContent)
 include(VersionFromGit)
 
 version_from_git()
@@ -19,6 +20,7 @@ elseif(CMAKE_SIZEOF_VOID_P EQUAL 4)
   message(STATUS "Building for 32 bit platform")
 endif()
 
+option(ENABLE_WINTLS_STANDALONE_ASIO "Enable Standalone WINTLS" OFF)
 option(ENABLE_TESTING "Enable Test Builds" ${WIN32})
 option(ENABLE_EXAMPLES "Enable Examples Builds" ${WIN32})
 option(ENABLE_DOCUMENTATION "Enable Documentation Builds" ${UNIX})
@@ -27,11 +29,35 @@ option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" ON)
 
 add_library(${PROJECT_NAME} INTERFACE)
 
+if(ENABLE_WINTLS_STANDALONE_ASIO)
+  FetchContent_Declare(
+    asio
+    GIT_REPOSITORY https://github.com/chriskohlhoff/asio.git
+    GIT_TAG        asio-1-28-0
+  )
+
+  FetchContent_GetProperties(asio)
+  if(NOT asio_POPULATED)
+    FetchContent_Populate(asio)
+  endif()
+endif()
+
 target_include_directories(${PROJECT_NAME}
   INTERFACE
   $<BUILD_INTERFACE:${${PROJECT_NAME}_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
+
+if(ENABLE_WINTLS_STANDALONE_ASIO)
+  target_include_directories(${PROJECT_NAME}
+    INTERFACE
+    ${asio_SOURCE_DIR}/asio/include
+  )
+  target_compile_definitions(${PROJECT_NAME}
+    INTERFACE
+    WINTLS_USE_STANDALONE_ASIO
+  )
+endif()
 
 target_compile_features(${PROJECT_NAME} INTERFACE cxx_std_14)
 
@@ -119,11 +145,13 @@ if(WIN32)
   set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${CMAKE_EXE_LINKER_FLAGS_RELEASE} /DEBUG /OPT:REF /OPT:ICF")
 endif()
 
-find_package(Boost REQUIRED)
+if(NOT ENABLE_WINTLS_STANDALONE_ASIO)
+  find_package(Boost REQUIRED)
 
-target_link_libraries(${PROJECT_NAME} INTERFACE
-  Boost::headers
-)
+  target_link_libraries(${PROJECT_NAME} INTERFACE
+    Boost::headers
+  )
+endif()
 
 if(MINGW)
   target_link_libraries(${PROJECT_NAME} INTERFACE
--- a/include/boost/wintls/certificate.hpp
+++ b/include/boost/wintls/certificate.hpp
@@ -11,18 +11,11 @@
 #include <boost/wintls/error.hpp>
 #include <boost/wintls/file_format.hpp>
 
-
 #include <boost/wintls/detail/config.hpp>
+#include <boost/wintls/detail/assert.hpp>
 #include <boost/wintls/detail/win32_crypto.hpp>
 
 #include <memory>
-#include <cassert>
-
-#if defined(NDEBUG)
-#define WINTLS_VERIFY_MSG(expr, msg) ((void)(expr))
-#else
-#define WINTLS_VERIFY_MSG(expr, msg) assert((expr) && (msg))
-#endif
 
 namespace boost {
 namespace wintls {
--- a/include/boost/wintls/certificate.hpp
+++ b/include/boost/wintls/certificate.hpp
@@ -11,9 +11,18 @@
 #include <boost/wintls/error.hpp>
 #include <boost/wintls/file_format.hpp>
 
+
+#include <boost/wintls/detail/config.hpp>
 #include <boost/wintls/detail/win32_crypto.hpp>
 
 #include <memory>
+#include <cassert>
+
+#if defined(NDEBUG)
+#define WINTLS_VERIFY_MSG(expr, msg) ((void)(expr))
+#else
+#define WINTLS_VERIFY_MSG(expr, msg) assert((expr) && (msg))
+#endif
 
 namespace boost {
 namespace wintls {
@@ -66,12 +75,12 @@ using cert_context_ptr = std::unique_ptr<const CERT_CONTEXT, detail::cert_ctx_de
  *
  * @return A managed cert_context.
  *
- * @throws boost::system::system_error Thrown on failure.
+ * @throws wintls::system_error Thrown on failure.
  *
  */
 inline cert_context_ptr x509_to_cert_context(const net::const_buffer& x509, file_format format) {
   // TODO: Support DER format
-  BOOST_VERIFY_MSG(format == file_format::pem, "Only PEM format currently implemented");
+  WINTLS_VERIFY_MSG(format == file_format::pem, "Only PEM format currently implemented");
 
   auto data = detail::crypt_string_to_binary(x509);
   auto cert = CertCreateCertificateContext(X509_ASN_ENCODING, data.data(), static_cast<DWORD>(data.size()));
@@ -96,11 +105,11 @@ inline cert_context_ptr x509_to_cert_context(const net::const_buffer& x509, file
  * @return A managed cert_context.
  *
  */
-inline cert_context_ptr x509_to_cert_context(const net::const_buffer& x509, file_format format, boost::system::error_code& ec) {
+inline cert_context_ptr x509_to_cert_context(const net::const_buffer& x509, file_format format, wintls::error_code& ec) {
   ec = {};
   try {
     return x509_to_cert_context(x509, format);
-  } catch (const boost::system::system_error& e) {
+  } catch (const wintls::system_error& e) {
     ec = e.code();
     return cert_context_ptr{};
   }
@@ -120,18 +129,18 @@ inline cert_context_ptr x509_to_cert_context(const net::const_buffer& x509, file
  *
  * @param name The name used to associate the key.
  *
- * @throws boost::system::system_error Thrown on failure.
+ * @throws wintls::system_error Thrown on failure.
  *
  * @note Currently only RSA keys are supported.
  */
 inline void import_private_key(const net::const_buffer& private_key, file_format format, const std::string& name) {
   // TODO: Handle ASN.1 DER format
-  BOOST_VERIFY_MSG(format == file_format::pem, "Only PEM format currently implemented");
+  WINTLS_VERIFY_MSG(format == file_format::pem, "Only PEM format currently implemented");
   auto data = detail::crypt_decode_object_ex(net::buffer(detail::crypt_string_to_binary(private_key)), PKCS_PRIVATE_KEY_INFO);
   auto private_key_info = reinterpret_cast<CRYPT_PRIVATE_KEY_INFO*>(data.data());
 
   // TODO: Set proper error code instead of asserting
-  BOOST_VERIFY_MSG(strcmp(private_key_info->Algorithm.pszObjId, szOID_RSA_RSA) == 0, "Only RSA keys supported");
+  WINTLS_VERIFY_MSG(strcmp(private_key_info->Algorithm.pszObjId, szOID_RSA_RSA) == 0, "Only RSA keys supported");
   auto rsa_private_key = detail::crypt_decode_object_ex(net::buffer(private_key_info->PrivateKey.pbData,
                                                                     private_key_info->PrivateKey.cbData),
                                                         PKCS_RSA_PRIVATE_KEY);
@@ -166,11 +175,11 @@ inline void import_private_key(const net::const_buffer& private_key, file_format
  *
  * @note Currently only RSA keys are supported.
  */
-inline void import_private_key(const net::const_buffer& private_key, file_format format, const std::string& name, boost::system::error_code& ec) {
+inline void import_private_key(const net::const_buffer& private_key, file_format format, const std::string& name, wintls::error_code& ec) {
   ec = {};
   try {
     import_private_key(private_key, format, name);
-  } catch (const boost::system::system_error& e) {
+  } catch (const wintls::system_error& e) {
     ec = e.code();
   }
 }
@@ -180,14 +189,14 @@ inline void import_private_key(const net::const_buffer& private_key, file_format
  *
  * @param name The name of the container storing the private key to delete.
  *
- * @throws boost::system::system_error Thrown on failure.
+ * @throws wintls::system_error Thrown on failure.
  *
  */
 inline void delete_private_key(const std::string& name) {
   HCRYPTKEY ptr = 0;
   if (!CryptAcquireContextA(&ptr, name.c_str(), nullptr, PROV_RSA_FULL, CRYPT_DELETEKEYSET)) {
 
-    throw boost::system::system_error(static_cast<int>(GetLastError()), boost::system::system_category());
+    throw wintls::system_error(static_cast<int>(GetLastError()), wintls::system_category());
   }
 }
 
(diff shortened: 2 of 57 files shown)


include/boost/wintls/certificate.hpp
```diff
@@ -11,9 +11,18 @@
...
+#include <boost/wintls/detail/config.hpp>
 #include <boost/wintls/detail/win32_crypto.hpp>
 
 #include <memory>
+#include <cassert>
+
+#if defined(NDEBUG)
+#define WINTLS_VERIFY_MSG(expr, msg) ((void)(expr))
+#else
+#define WINTLS_VERIFY_MSG(expr, msg) assert((expr) && (msg))
```
laudrup, line 24: This should be moved to a different header. Not sure if one of the existing ones would make sense or a new one should be added though.

include/boost/wintls/detail/async_handshake.hpp
```diff
@@ -10,16 +10,21 @@
...
 #include <boost/wintls/handshake_type.hpp>
 
+#include <boost/wintls/detail/config.hpp>
 #include <boost/wintls/detail/sspi_handshake.hpp>
 
+#ifdef WINTLS_USE_STANDALONE_ASIO
+#include <asio/coroutine.hpp>
+#else
 #include <boost/asio/coroutine.hpp>
+#endif
```
laudrup, line 20: I think it might make more sense to move this to `detail/config.hpp` or perhaps even better, introduce a `detail/coroutine.hpp` header that includes the correct header as well as defines the related macros.

include/boost/wintls/detail/sspi_handshake.hpp
```diff
@@ -70,7 +76,7 @@ class sspi_handshake {
         case handshake_type::server:
           return SECPKG_CRED_INBOUND;
       }
-      BOOST_UNREACHABLE_RETURN(0);
+      std::abort(); // unreachable here
```
laudrup, line 79: This doesn't really do the same as the boost macro.

Might make sense to move this in another header that calls the correct builtin for various compilers. Maybe the same place as the assert macro?

I'm not a huge of generic names like `util.hpp` but in this case it might be acceptable.

Alternatively add separate headers for various compilers included from `config.hpp` similar to what boost does.

include/boost/wintls/detail/sspi_handshake.hpp
```diff
@@ -234,18 +240,19 @@ class sspi_handshake {
...
 
       default:
         return state::error;
     }
   }
 
   void size_written(std::size_t size) {
-    BOOST_VERIFY(size == out_buffer_.size());
+    (void)(size);
+    assert(size == out_buffer_.size());
```
laudrup, line 255: If the `BOOST_VERIFY_MSG` macro is reimplemented, might as well reimplement both of them.

laudrup (commented):
Looks really good.

As far as I can tell it is mostly just `s/boost::system/wintls/g` like changes which is great as it is even simpler than I had hoped. That doesn't mean I don't appreciate your effort, quite the opposite :-)

I have a few comments related to moving some things to headers that could probably be general comments as I just commented the first time I saw it but I hope you get the point. These are hopefully minor things that should be relatively trivial to fix.

The one major thing I would like though, is to have this build and hopefully also tested in the build pipeline (github actions). Not really sure which kind of combinations would make sense to test but I would appreciate it if you could try to see if you could add that as well.

Then of course an update to README is needed and maybe even changing the name as this would probably be more aptly named `asio.wintls` instead of `boost.wintls` :-D

Thanks a lot for this. Could potentially greatly improve the target audience for this library.

windowsair:
Thank you for your review. It seems that our thinking is feasible, it's so good!  I saw your comment and it was so valuable. But now, let's consider the test code first.

I'm not sure what to do with the test code, so I'll stick with the basic idea of this pr, which is to remove as many relevant boost dependencies as possible. Notice that the main dependency in the test code is beast, and we use `stream.hpp` from the beast test code. Unfortunately, beast's test code also depends on some of its own basic components.

Fortunately, rewriting them wasn't as hard as I thought it would be, and after rewriting this part of the file, we can get a `stream.hpp` that is basically free of boost dependencies (but there are still a few dependencies, such as some utility templates, which are features provided in the C++20 STL).

The added code structure is shown below:
```bash
└── test
    ├── utils
    │   ├── allocator.hpp
    │   ├── empty_value.hpp
    │   ├── error.hpp
    │   ├── fail_count.hpp
    │   ├── flat_buffer.hpp
    │   ├── impl
    │   │   ├── error.ipp
    │   │   ├── fail_count.ipp
    │   │   ├── flat_buffer.hpp
    │   │   ├── stream.hpp
    │   │   └── stream.ipp
    │   ├── stream.hpp
    │   └── test_config.hpp
```

As mentioned just now, this part of the code comes from Beast, with some additional modifications. This part of the code totals about 2k lines and can be used to replace the original dependency `stream.hpp`.  Now we can then use the namespace alias trick again to distinguish whether to use standalone-wintls or boost-wintls for testing.

Now, I've made a first attempt and passed unittest on the boost version. I can push these commit together. Or if you have a better idea, please let me know, thank you!

windowsair:
Yes, in effect what we've done is ported a minimal beast test component to our test code so that we don't depend on beast.  And then we can use this "minimal beast" to build our test code. I think the option "whether to use beast to build test code" should no longer be needed here.

If that's appropriate, I'll try to organize some commits and submit them :-D

laudrup:
@windowsair That sounds great. Looking forward to see the resulting commits.

laudrup:
Thanks a lot for this. Looks really good.

Some general questions/comments:

I'm not really sure I understand which parts of the test code requires either boost or C++17 apart from the Boost.Process part. I would very much like us to not depend on boost at all for this to make sense and as far as I can tell that should be possible.

I initially chose to target C++14 since that is the first C++ standard that Microsofts compiler clams to support. I would rather not change that unless there are some good reasons. It could potentially limit the amount of users able to use this but of course I have no idea how relevant that is. As far as I can tell, the compilers actually being used for testing this code all support C++17 well enough, but if we should target C++17 I would prefer to keep that in a separate discussion.

Regarding Boost.Process. It seems like it is only being used for starting and stopping a process with some command line arguments. Although the WIN32 API pretty terrible it should be relatively simple to write the required code for calling `StartProcessEx` and other functions to do the same. It doesn't seem like we need to do anything like redirecting streams and we even only have to support one platform.

If you don't feel like it, I might see if I can find the time to implement something to get rid of the Boost.Process dependency.

Thanks once more. This is definitely making good progress. And yes, I can have a look at updating the CI once everything else is in place.

windowsair:
Ah, I actually don't dislike Boost. At first I just wanted to get rid of the Beast dependency, because standalone-wintls couldn't work well with it. Hmmm, since you want the standalone version to not rely on Boost at all, it looks like we have more work to do.

Regarding C++17, the feature used is `string_view`, you can see the details in `stream.hpp`:
```cpp
#if __cplusplus >= 201703L || (defined _MSVC_LANG && _MSVC_LANG >= 201703L)
#include <string_view>
using string_view = std::string_view;
#else
#include <boost/utility/string_view.hpp>
using string_view = boost::string_view;
#endif
```

Some member functions of the stream class use `string_view` . I'm not sure it's a good idea to just switch to `std::string`.

Regarding Boost.Process, I use this in my own projects: https://github.com/sheredom/subprocess.h

Quite simple header-only subprocess library. With it, I can replace Boost.Process in a few minutes. Of course, if you don't like to include external dependencies, I think this is still a good reference for implementing subprocess.

windowsair:
At first I introduced the standalone Beast test just to test standalone-wintls. As you know, Beast uses Boost-asio, while standalone-wintls uses non-Boost-Asio, which may lead to potential problems. Completely removing the Boost part of the test code was not what I originally intended. Of course if it can be done easily, why not do that 😄

I totally agree with you about `string_view` and Boost.Process. Also, for standalone-Asio, we might need to handle it via CMake FetchContent as well.

What is our next move? Is there any way I can help? Thank You.

laudrup:
If you want to try to make the tests work without depending on boost at all that would be great. Otherwise I can look into that at some point.

As far as I can tell, there are some minor issues that I've commented on that it doesn't look like you've addressed so if that is the indeed the case then please try to fix those.

Otherwise it is fine with me if I hijack this branch and do what is needed to extend the pipeline, update the documentation and probably even changing the name of this project/library.

I'll of course make sure you get the credit you deserve but it's up to you really how much more time you want to spent on this.

windowsair:
Okay, let's see how the code is performing so far.

The first patch is about the main code changes, which were made based on your comments. The other patches are some work to replace Boost, and now we really don't need any Boost code in our test code for standalone version.

The next step might be to add standalone-asio in Cmake, but before that let's see if there are any improvements to be made in these patches :D

jens-diewald:
Hi @laudrup,
hi @windowsair,

If I may voice an opinion from an interested observer:
I think the original effort from @windowsair is great. Making this library useful for people that do not want to depend on all of boost is probably a good idea.
But the additional efforts of removing all boost dependencies from the test code look way over the top for me.
Effectively forking other (large!) 3rd Party Libs into this Repository does not seem like a good idea to me, when we can just use boost..?

windowsair:
Hi, @jens-diewald ,

Thank you for your attention. As I said, it should work fine even if we don't remove all the Boost parts from the test code.
What I did in the last few patches in fact removed these few Boost dependencies:

- `Boost.Process`: replaced by `subprocess.h`, about 1k lines code, single file.
- `string_view`: replaced by `string_view-lite`, about 1700 lines code, single file.
- `variant`: replaced by `variant-lite`, about 2800 lines code, single file.

If C++17 was available, that would be great, we could just use `std::string_view` and `std::variant`, but we chose C++14 as the target, so we can only replace these by bringing in 3rd libraries. It seems to me that it might be of some use to those who don't want to use Boost at all. 

I'm not sure if this is too large. Maybe I misunderstood what you meant?

laudrup:
@windowsair First of all, thanks a lot for your great efforts on this PR.

Overall, it looks good now. One major thing though, I'd really prefer that we use CMake `FetchContent` for handling the external dependencies instead of copying them to this repository directly.

@jens-diewald Of course your input it very much appreciated. I don't have any use for this library myself (at least currently) so I appreciate all the input I can get.

So regarding your comment and some open questions to both of you whom I assume are interested in using this library and most likely have more experience with Windows than I do.

I assumed the only thing we really need from boost for the tests of this library is the test stream in `boost.beast`. Since that is more or less an internal thing used by the tests of Beast itself I think it can make good sense to copy our own version to this library. Vinnie from the Beast library has more or less encouraged me to do exactly that and it also makes it easier for change or extend the test stream if we would want that. I don't think that is at all unlikely.

The only other thing we actually use from boost is the process library. Since we only use that to start and stop a process, I think it's a bit overkill to pull in an external dependency just for that. It makes sense to use `boost.process` if we already have boost as a dependency, but writing a simple class that starts a process and stops it on destruction ought to be fairly trivial even if it involves the nastiness of the Win32 API.

@windowsair chose to pull in an external dependency for that which is fine, but I had definitely considered getting rid of that at some point and write the process wrapper code at some later point.

A bit to my surprise it turned out that we actually depend on either C++17 or boost for `variant` and `string_view` implementations, which leads me to a related but somewhat different issue.

I cannot remember exactly why I chose to target C++14 for this library initially, but as far as I remember I tried to make it easy for someone already using `boost.asio` to use this library without any additional requirements on compiler, toolchain etc. Since MSVC only started to be somewhat standards compliant after C++11 and only started supported setting a standards flag after C++14, that's what I aimed for. I'm fairly certain that `asio` requires at least C++11 so it was an effort not to divert to much from that.

On the other hand, the only compilers/toolchains we actually test with in our CI pipeline all supports C++17 just fine so it might make sense to drop that requirement and instead of supporting a specific C++ standard simply tests against a set of compilers which is more or less what is already being done anyway.

I wouldn't mind dropping the C++14 support and simply document that VS 2017 and VS 2019 is tested and supported while earlier versions might work. According to [this](https://learn.microsoft.com/en-us/visualstudio/productinfo/vs-servicing) official support for VS 2013 stops early next year so the question is really if it is worth the effort to support VS 2015? That version is not available in the Github actions anyway so it's not like it's even been tested.

I would love to hear you guys opinion on that. It would fit me just fine to drop the C++14 support, use `std::variant` and `std::string_view` for the tests. If someone really needs support for older versions they are free to open an issue or a pull request.

Please let me know what you think. If you think it sound reasonable I'll drop the C++14 flag and update the master branch so we can rebase this PR and hopefully get this merged soon.

Thanks a lot for your contributions to this project. I very much appreciate it.

windowsair:
Hi, @laudrup

Thank you for your guidance. I think breaking C++14 support might not be a good idea.

Although the CI already supports new enough compiler versions, but we can't be sure what environment the user is using. Perhaps they need the unit test programs we provide to determine if their current compiler and system environment will run the code properly.

Importing external libraries as `FetchContent` is easy to do, and I want to focus on the subprocess library here:

The functions we currently need are probably to start a process, terminate a process, and determine whether a process is alive or not. I'm not sure we'll need to read stdin/stdout and other functions in the future. The current 3rd libraries basically call some Win32API. Of course there is also some handling of process parameters, posix support and handling of imported functions (no include <windows.h>), and different compiler handling operations.

These parts are not complicated, but they are quite trivial. Third party libraries have taken care of all of this with only about 1000 lines code, which doesn't seem particularly large to me. This is why I tend to use 3rd libraries rather than implementing them myself.

I've prepared the relevant changes regarding the standalone-asio Cmake test. Once we reach a consensus, I will commit it ASAP.

jens-diewald:
Hi @laudrup and @windowsair,

sorry for the late reply, I did not find any time to respond during the holidays.

Regarding my own requirements for this library: I would be fine with C++20 and all of Boost.
But I am aware that not everyone is in that position. I do suspect, C++14 is still a reasonable requirement today.
@laudrup, what are you trying to get at regarding the end of support for VS 2013? C++14 is still supported (and the default) in VS 2022 and the GitHub pipeline only tests VS 2019 and VS 2022. Or am I missing something?

Also, I think that it is a reasonable assumption, that there are people who would be happy using Standalone WinTLS+Asio and expanding the group of potential users is great!

However, I do not think that it is anywhere near as important, to have the tests not depend on boost.
@windowsair raises the point, that even users may want to run the tests. I did not think of that, and I still do not really think that most (all?) users would want to do that. They should write their own tests for whatever-they-use-wintls-for.
That means, removing the Boost dependency from the tests would only be a simplification for contributors to this library that use the standalone version otherwise. (Also note that the tests depend on OpenSSL as well.)
On the other side - and that may be an opinion you do not share - I fear that introducing more external dependencies will make maintaining this library significantly more work in the long run, while there are only so many man-hours available even now.
So, in my opinion, having our own process wrapper/string_view/variant, be it another non-boost library or self-written, costs time that could better be used for something else.
Regarding the Beast test stream, I do not think that it is possible to reasonably use that with standalone WinTLS+Asio. I do think though, it may be fine to completely skip all tests using that stream for the standalone version. However, depending on the Beast test stream is a little bit awkward by itself. It would be nicer if asio would provide such a stream as an "official" testing facility by itself. So, I am not that opposed to forking the Beast test stream.

So, all in all, I think removing the Boost dependency from the tests costs time better used elsewhere.
If you choose to use the additional libraries, I agree that using `FetchContent` would be nicer that forking them into the repository.

laudrup:
Hi @windowsair,

Thanks a lot for you work on this and to @jens-diewald for the very valuable input.

I have tried to [github actions a bit](https://github.com/laudrup/boost-wintls/tree/github-actions-updates) to test building with different C++ standards. In order to not get an unreasonable amount of test jobs that also means cutting down a bit on which versions of boost to test against. This was meant as a preparation for adding tests for both standalone asio and boost.asio.

I think it's about time I get this merged. While we can spend lots of time continuing discussing the pros and cons of thirdparty dependencies etc. I think it's better to get this merged to the master branch and tagged with a version so we can get this out there and potentially get some feedback.

So, @windowsair if you commit your latest changes I think the best way forward is for me to hijack this branch, remove the boost references from the name(space) and other updates so this will become `asio-wintls` instead. It might also involve some other minor changes to the changes you already made but I hope you're OK with that.

Then we can finally close this PR. Are you OK with that?

windowsair:
Hi, @laudrup ,

Sure. Let's start merging this!

windowsair:
Hi, @laudrup 

The new patches add option `ENABLE_STANDALOND_WINTLS` to determine whether to enable the standalone version of wintls compilation.

There is also a new change in CI:

Using `choco install openssl` may fetch different versions of OpenSSL, one being 3.1.1 and the other being 3.2.0, which causes failures in our test cases.

I think this is probably not caused by the changes in this PR. Using the code from the master branch and explicitly specifying OpenSSL version as 3.2.0 will also result in failures for all test cases. See:

https://github.com/windowsair/boost-wintls/tree/openssl-test
https://github.com/windowsair/boost-wintls/actions/runs/7528290666

Openssl version 3.2.0 was recently introduced by choco. It looks like choco in CI randomly installs both version 3.1.1 and 3.2.0, and once it installs version 3.2.0, then our test case fails.  So another patch specified Openssl as version 3.1.1 to bypass this issue. We may need to investigate this issue in the future.

I think there are probably no more commits to push now. Feel free to modify them :)

CMakeLists.txt
```diff
@@ -27,12 +29,36 @@ option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" ON)
 
 add_library(${PROJECT_NAME} INTERFACE)
 
+if(ENABLE_STANDALOND_WINTLS)
+  FetchContent_Declare(
+    asio
+    GIT_REPOSITORY https://github.com/chriskohlhoff/asio.git
+    GIT_TAG        asio-1-28-0
+  )
```
Nerixyz, line 37: Should people really opt-in to using `FetchContent`? I think it would be better if users would provide ASIO by themselves (e.g. some users prefer using git submodules).

CMakeLists.txt
```diff
@@ -19,6 +20,7 @@ elseif(CMAKE_SIZEOF_VOID_P EQUAL 4)
   message(STATUS "Building for 32 bit platform")
 endif()
 
+option(ENABLE_STANDALOND_WINTLS "Enable Standalone WINTLS" OFF)
```
Nerixyz, line 23:
```suggestion
option(ENABLE_STANDALONE_WINTLS "Enable Standalone WINTLS" OFF)
```

There's a typo here. I don't think there's any naming standard for CMake options, however I often see the options prefixed by the project name (like for #defines). So it could be something like `WINTLS_STANDALONE_ASIO` here.

CMakeLists.txt
windowsair: That might lose the out-of-the-box experience. If the user has such a need, I think they should modify the cmake file. Furthermore, WINTLS is a header-only lib, some users may not even use the cmake that comes with the project.

laudrup: I agree with @Nerixyz. I don't think it makes sense for this library to fetch Asio with CMake, as a git submodule any other way.

Instead, we should create a `FindAsio.cmake` file and use that similar to how Boost is found.

windowsair: Okay, I'll spend some time working on this. Or you can hijack this branch to modify them.

laudrup: @windowsair

I can takeover from here and do the last things needed for this to be merged to the master branch, OK?

windowsair: @laudrup 

Of course :D

laudrup:
@windowsair 

I've picked up your branch and spent quite some time adding new commits. Some related to handle thirdparty code differently and mainly to rename this library from `boost-wintls` to simply `wintls`.

My plan is pretty much to merge [this branch](https://github.com/laudrup/boost-wintls/tree/wintls-standalone) as it is and then rename this repository to `wintls`.

I think you should have a chance to have a look and comment before I do this. Of course your commits will still be in the history mostly unchanged (I've fixed a few conflicts I think). If you want more credit let me know.

@jens-diewald Your input would also be appreciated if you have any.

Thanks!

windowsair:
Hi, @laudrup 

Looks really good. Lets see what others say.

laudrup:
I have merged these changes to master so I will close this pull request.

Thanks for you contribution @windowsair

Issue #84, opened 2024-05-10
Fails to compile with UNICODE defined


computerquip-work:
```
inline SECURITY_STATUS AcquireCredentialsHandle(SEC_CHAR* pPrincipal,
                                                SEC_CHAR* pPackage,
                                                unsigned long fCredentialUse,
                                                void* pvLogonId,
                                                void* pAuthData,
                                                SEC_GET_KEY_FN pGetKeyFn,
                                                void* pvGetKeyArgument,
                                                PCredHandle phCredential,
                                                PTimeStamp ptsExpiry)
```
If the build system defines UNICODE, AcquireCredentialsHandle should expand to AcquireCredentialsHandleW where AcquireCredentialHandle uses SEC_WCHAR instead of SEC_CHAR. Thus, the following call will always fail to compile:
```
  return sspi_function_table()->AcquireCredentialsHandle(pPrincipal,
                                                          pPackage,
                                                          fCredentialUse,
                                                          pvLogonId,
                                                          pAuthData,
                                                          pGetKeyFn,
                                                          pvGetKeyArgument,
                                                          phCredential,
                                                          ptsExpiry);
```

A similar problem exists with `InitializeSecurityContext` as well. It might be necessary make some sort of `SEC_TCHAR` that defines as either `SEC_WCHAR` or `SEC_CHAR` based on if `UNICODE` is defined. Alternatively, instead of using the generic name that changes based on `UNICODE`, could just use the name wanted directly, i.e. `AcquireCredentialsHandleA`.

laudrup:
I have not been able to reproduce this issue.

The way the SSPI headers are included [here](https://github.com/laudrup/boost-wintls/blob/0077d733984f84a2a85e216ced29f73b173ae472/include/wintls/detail/sspi_types.hpp#L16)  should ensure that the non wide-character versions of the functions are being used.

[This job](https://github.com/laudrup/boost-wintls/blob/0077d733984f84a2a85e216ced29f73b173ae472/.github/workflows/unittest.yml#L41) tests building the code with `UNICODE` defined, so I'm fairly confident that it works.

It might not be that best way to handle it but it would be nice with some more information before I look more into this.

Thanks.

computerquip-work:
I'd have to take another look at it but I'd wager it's because those headers are included before wintls is included so it doesn't get the chance to define the types wintls wants.

laudrup:
[...]

PR #85, opened 2024-05-17
Use ANSI functions explicitly

computerquip-work:
A simple attempt at fixing the issues with UNICODE. I've not removed the countermeasures already implemented.

Work towards: https://github.com/laudrup/boost-wintls/issues/84


laudrup:
Looks like how this should have been done initially. Thanks a lot.

If you remove the ugly `#ifdef UNICODE` stuff I had written initially I think this could be merged.

There's already a job for testing with `UNICODE` defined. You might consider extending that so it fails with the current implementation (ie. the headers are included with `UNICODE` defined before including wintls) but it's not all that important.

laudrup:
Sorry,  I had completely forgotten about this

Issue #83, opened 2024-02-25
TLS v1.3 not work


windowsair:
It looks like TLS v1.3 is not working properly, as shown in the unit test.

I think we should change the API we used, as shown in the following reference link:
https://github.com/curl/curl/pull/7784

Unfortunately, it doesn't appear that MS has documented this.

laudrup:
Hi @windowsair 

Sorry for the very late reply.

It seems like you are right that this does not work when building on Windows 2022 where TLS 1.3 ought to be supported.

I am not sure what might be the cause of this but I am afraid I'm not really able to look into this as I don't currently have access to any build machine running Windows 2022 (or Windows 11) which seems to be required to use TLS 1.3 with SChannel.

Implementing by pushing and getting output from GitHub actions is not really worth it, so unless I get access to a Windows 11 installation somehow I don't think I can fix this right now.

Any pull requests related to this would be very much appreciated.

PR #86, opened 2024-07-07
Add support for TLS v1.3

windowsair:
TLS v1.3 is available on windows 10.0.17763 and above version. #83 

The associated unittest needs further modification, I'll upload it later.

### DIFF
--- a/include/wintls/detail/sspi_handshake.hpp
+++ b/include/wintls/detail/sspi_handshake.hpp
@@ -47,17 +47,10 @@ public:
   void operator()(handshake_type type) {
     handshake_type_ = type;
 
+    TLS_PARAMETERS tls_parameters{};
+    SCH_CREDENTIALS credentials{};
     SCHANNEL_CRED creds{};
-    creds.dwVersion = SCHANNEL_CRED_VERSION;
-    creds.grbitEnabledProtocols = static_cast<DWORD>(context_.method_);
-    creds.dwFlags = SCH_CRED_MANUAL_CRED_VALIDATION | SCH_CRED_NO_DEFAULT_CREDS;
-    // If revocation checking is enables, specify SCH_CRED_REVOCATION_CHECK_CHAIN_EXCLUDE_ROOT
-    // to cause the TLS certificate status request extension (commonly known as OCSP stapling)
-    // to be sent. This flag matches the CERT_CHAIN_REVOCATION_CHECK_CHAIN_EXCLUDE_ROOT
-    // flag that we pass to the CertGetCertificateChain calls during our manual authentication.
-    if (check_revocation_) {
-      creds.dwFlags |= SCH_CRED_REVOCATION_CHECK_CHAIN_EXCLUDE_ROOT;
-    }
+    void* cred = nullptr;
 
     auto usage = [this]() {
       switch (handshake_type_) {
@@ -69,13 +62,63 @@ public:
       WINTLS_UNREACHABLE_RETURN(0);
     }();
 
+    auto server_cert = context_.server_cert();
+    bool is_tlsv13 = [this]() {
+      switch (context_.method_) {
+        case method::tlsv13:
+        case method::tlsv13_client:
+        case method::tlsv13_server:
+          return true;
+        default:
+          return false;
+      }
+      WINTLS_UNREACHABLE_RETURN(0);
+    }();
+
+    DWORD version = is_tlsv13 ? SCH_CREDENTIALS_VERSION : SCHANNEL_CRED_VERSION;
+    DWORD flags = is_tlsv13 ? SCH_USE_STRONG_CRYPTO : (SCH_CRED_MANUAL_CRED_VALIDATION | SCH_CRED_NO_DEFAULT_CREDS);
+    DWORD protocols = static_cast<DWORD>(context_.method_);
+
+    // If revocation checking is enables, specify SCH_CRED_REVOCATION_CHECK_CHAIN_EXCLUDE_ROOT
+    // to cause the TLS certificate status request extension (commonly known as OCSP stapling)
+    // to be sent. This flag matches the CERT_CHAIN_REVOCATION_CHECK_CHAIN_EXCLUDE_ROOT
+    // flag that we pass to the CertGetCertificateChain calls during our manual authentication.
+    if (check_revocation_) {
+      flags |= SCH_CRED_REVOCATION_CHECK_CHAIN_EXCLUDE_ROOT;
+    }
+
+    DWORD num_creds = 0;
+    decltype(&server_cert) creds_list = nullptr;
+
+    if (handshake_type_ == handshake_type::server && server_cert != nullptr) {
+      num_creds = 1;
+      creds_list = &server_cert;
+    }
+
     // TODO: rename server_cert field since it is also used for client cert.
     // Note: if client cert is set, sspi will auto validate server cert with it.
     // Even though verify_server_certificate_ in context is set to false.
-    auto server_cert = context_.server_cert();
-    if (server_cert != nullptr) {
-      creds.cCreds = 1;
-      creds.paCred = &server_cert;
+    if (handshake_type_ == handshake_type::client && server_cert != nullptr) {
+      num_creds = 1;
+      creds_list = &server_cert;
+    }
+
+    if (!is_tlsv13) {
+      cred = &creds;
+      creds.dwVersion = version;
+      creds.grbitEnabledProtocols = protocols;
+      creds.dwFlags = flags;
+      creds.cCreds = num_creds;
+      creds.paCred = creds_list;
+    } else {
+      cred = &credentials;
+      credentials.dwVersion = version;
+      credentials.dwFlags = flags;
+      credentials.cTlsParameters = 1;
+      credentials.pTlsParameters = &tls_parameters;
+      credentials.pTlsParameters->grbitDisabledProtocols = ~protocols;
+      credentials.cCreds = num_creds;
+      credentials.paCred = creds_list;
     }
 
     TimeStamp expiry;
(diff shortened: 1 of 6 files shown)


windowsair:
Okay, now it is ready to review :)

include/wintls/detail/sspi_handshake.hpp
```diff
@@ -19,13 +19,61 @@
...
+
+    DWORD               cMappers;
+    struct _HMAPPER **aphMappers;
+
+    DWORD               dwSessionLifespan;
+    DWORD               dwFlags;
+    DWORD               cTlsParameters;
+    PTLS_PARAMETERS     pTlsParameters;
+} SCH_CREDENTIALS, * PSCH_CREDENTIALS;
+
```
laudrup, line 76: Why are all these C types needed here? Aren't they already defined in some of the SSPI headers already included.

If they are really needed they definitely shouldn't be here.

include/wintls/detail/sspi_handshake.hpp
```diff
@@ -74,25 +115,67 @@ class sspi_handshake {
     }();
 
     auto server_cert = context_.server_cert();
-    if (handshake_type_ == handshake_type::server && server_cert != nullptr) {
-      creds.cCreds = 1;
-      creds.paCred = &server_cert;
-    }
+    bool is_tlsv13 = !!(static_cast<int>(context_.method_) & static_cast<int>(method::tlsv13));
```
laudrup, line 118: I'm not really a fan of this C type code as it depends on the underlying value of the enum and is a bit hard to read. Please consider using an inline lambda similar to what is being done for the `usage` variable above.

include/wintls/detail/sspi_handshake.hpp
```diff
@@ -51,17 +99,10 @@ class sspi_handshake {
...
-    creds.grbitEnabledProtocols = static_cast<DWORD>(context_.method_);
-    creds.dwFlags = SCH_CRED_MANUAL_CRED_VALIDATION | SCH_CRED_NO_DEFAULT_CREDS;
-    // If revocation checking is enables, specify SCH_CRED_REVOCATION_CHECK_CHAIN_EXCLUDE_ROOT
-    // to cause the TLS certificate status request extension (commonly known as OCSP stapling)
-    // to be sent. This flag matches the CERT_CHAIN_REVOCATION_CHECK_CHAIN_EXCLUDE_ROOT
-    // flag that we pass to the CertGetCertificateChain calls during our manual authentication.
-    if (check_revocation_) {
-      creds.dwFlags |= SCH_CRED_REVOCATION_CHECK_CHAIN_EXCLUDE_ROOT;
-    }
+    void *pCred;
```
laudrup, line 105: Please avoid hungarian notation and please initialize variables.

include/wintls/detail/sspi_handshake.hpp
```diff
@@ -74,25 +115,67 @@ class sspi_handshake {
...
-      creds.cCreds = 1;
-      creds.paCred = &server_cert;
+      // TODO: rename server_cert field since it is also used for client cert.
+      // Note: if client cert is set, sspi will auto validate server cert with it.
+      // Even though verify_server_certificate_ in context is set to false.
+      if (handshake_type_ == handshake_type::client && server_cert != nullptr) {
+        credentials.cCreds = 1;
+        credentials.paCred = &server_cert;
+      }
     }
```
laudrup, line 171: Most of the code is duplicated in the different if/else branches here (`is_tlsv3`).

Please find a way to refactor this to avoid that duplication.

test/tls_record.cpp
```diff
@@ -77,8 +84,78 @@ tls_handshake::message_type read_message(tls_handshake::handshake_type t, net::c
...
+      return tls_extension::common{buffer, size};
+    case tls_extension::extension_type::post_handshake_auth:
+      return tls_extension::common{buffer, size};
+    case tls_extension::extension_type::signature_algorithms_cert:
+      return tls_extension::common{buffer, size};
+    case tls_extension::extension_type::key_share:
+      return tls_extension::common{buffer, size};
+    default:
+      return tls_extension::common{buffer, size};
+  }
```
laudrup, line 137: I might be reading this the wrong way, but it definitely looks like this could more or less be reduced to:

`return tls_extension::common{buffer, size};`

Also, if you are going to switch on all the enum values, then please leave out the default case. That way the compiler can warn if an enum value isn't handled which is definitely nice.

So maybe this could be written as:

```
 switch (t) {
    case tls_extension::extension_type::server_name:
    case tls_extension::extension_type::max_fragment_length:
    case tls_extension::extension_type::status_request:
    // Remaing enum values follow...
     return tls_extension::common{buffer, size};
 }
WINTLS_UNREACHABLE_RETURN(0);
```

laudrup (changes requested):
Thanks a lot for this very useful improvement.

Very good job an extending the unit tests so they also test that this actually does handle TLS v1.3 (I assume it does so :-)).

Please address my comments though.

If you don't have the time or don't want to make these changes yourself then I'll be happy to do so myself when time permits. Your contribution is much appreciated no matter what.

include/wintls/detail/sspi_handshake.hpp
```diff
@@ -19,13 +19,61 @@
 
 #include <wintls/handshake_type.hpp>
 
+#include <SubAuth.h>
```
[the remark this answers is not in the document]
windowsair: struct member `UNICODE_STRING` in `CRYPTO_SETTINGS` need it.

> To use the SCH_CREDENTIALS structure, define SCHANNEL_USE_BLACKLISTS along with UNICODE_STRING and PUNICODE_STRING. Alternatively, include Ntdef.h, SubAuth.h or Winternl.h.

include/wintls/detail/sspi_handshake.hpp
windowsair: > Minimum supported client	Windows 10 1809 [desktop apps only]

Doing so helps us to compile on lower versions of Windows.

include/wintls/detail/sspi_handshake.hpp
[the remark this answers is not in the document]
laudrup: OK. Then please move it so it's included with the other SSPI related headers.

include/wintls/detail/sspi_handshake.hpp
laudrup: > > Minimum supported client	Windows 10 1809 [desktop apps only]
> 
> Doing so helps us to compile on lower versions of Windows.

OK. Then please move it to it's own header and only include if compiling on a version of Windows where it's required. This should probably be in `sspi_types.h`.

windowsair:
There are conflicting codes. But the code in mainline doesn't seem to pass CI. Do I need to wait for the mainline code to be fixed first?

laudrup:
> There are conflicting codes. But the code in mainline doesn't seem to pass CI. Do I need to wait for the mainline code to be fixed first?

As far as I can tell the errors were related to a problem with the Windows images used by all Github CI actions, not just this repository, so you should be able to fix the conflicts and rebase without any issues now.

windowsair:
> > There are conflicting codes. But the code in mainline doesn't seem to pass CI. Do I need to wait for the mainline code to be fixed first?
> 
> As far as I can tell the errors were related to a problem with the Windows images used by all Github CI actions, not just this repository, so you should be able to fix the conflicts and rebase without any issues now.

Unit test works fine. But coverage and coverity task failed :(

laudrup:
> Unit test works fine. But coverage and coverity task failed :(

The coverage and coverity jobs require access tokens which are only available on my github repository. I assume you could generate your own accounts on codecov and coverity if you really wanted to, but that'd hardly be worth it.

Just ignore that for now and please focus on addressing the issues I've pointed out instead.

The coverage and coverity jobs are only being run on the master branch anyway so they won't stop this PR from being merged.

windowsair:
> > Unit test works fine. But coverage and coverity task failed :(
> 
> The coverage and coverity jobs require access tokens which are only available on my github repository. I assume you could generate your own accounts on codecov and coverity if you really wanted to, but that'd hardly be worth it.
> 
> Just ignore that for now and please focus on addressing the issues I've pointed out instead.
> 
> The coverage and coverity jobs are only being run on the master branch anyway so they won't stop this PR from being merged.

Okay, it;s ready for review

include/wintls/detail/sspi_types.hpp
```diff
@@ -29,4 +29,51 @@
 #define UNICODE
 #endif // WINTLS_UNICODE_UNDEFINED
 
+#include <SubAuth.h>
```
laudrup, line 32: Please move this up between the `#ifdefs` where the two other includes are. Hopefully that doesn't cause any issues. Otherwise we should look into that.

include/wintls/detail/sspi_types.hpp
```diff
@@ -29,4 +29,51 @@
...
+
+    DWORD               cMappers;
+    struct _HMAPPER **aphMappers;
+
+    DWORD               dwSessionLifespan;
+    DWORD               dwFlags;
+    DWORD               cTlsParameters;
+    PTLS_PARAMETERS     pTlsParameters;
+} SCH_CREDENTIALS, * PSCH_CREDENTIALS;
+
```
laudrup, line 78: You mentioned that these structs only needed to be defined when building on some older versions of Windows.

Could you please move them to an appropriately named header file and then only include that file if we are building on a version of Windows where they are needed?

laudrup (changes requested):
Sorry for being late with this review.

It looks very good. Please address my last comments and then I think this could be merged.

I also made some changes to the master branch so the coverity and coverage jobs should no runner long on forks of this repository. You made me aware of that issue so thanks a lot for that as well. You might want to rebase your fork on this repos master branch.

include/wintls/detail/sspi_types.hpp
windowsair: Already moved to `sspi_compat.hpp`

laudrup:
Looks great.

Only one minor comment left then I think should be ready to merge and I should probably also make a new release with this.

Thanks a lot.

laudrup:
@windowsair 

Seems like I forgot to actually submit my final comment. After that I think we should merge this.

Thanks.

laudrup:
Great and thanks a lot.

Sorry for the slow response time.

PR #89, opened 2024-07-30
Move manual_auth out of the handshake loop.

jens-diewald:
Cherry picked from #77.
This was the first proper change from that PR.
I will wait for this to be reviewed and hopefully merged before getting the other changes from that PR up to date.

Short explanation: There was some redundancy in the handling of `sspi_handshake::state::done` and `sspi_handshake::state::done_with_data` and the corresponding `error` values.
This changes the code such that the manual auth is only called after `sspi_handshake::state::done` was produced, removing the need for the two `*_with_data` values.


laudrup:
Removing code and simplifying things is always great :+1: 

Thanks a lot for this.

Issue #91, opened 2025-03-10
TLS 1.3 handshake issues


mkentie:
I'm trying to use boost-wintls with a simple webserver based on the asio single-threaded webserver example.
In tls 1.2 mode this works fine. In tls 1.3 mode, this hangs the server if I use async handshaking; sync handshaking works fine.
In an effort to simplify this problem, I've tried to modify the echo server/client to use tls 1.3. However, those now fail with both sync and async handshaking. Server says:

Handshake failed: The specified data could not be decrypted

Client says:

Read failed: End of file

All I did for both examples was change `wintls::method::system_default` to `tlsv13_server` and client. Changing this to tlsv12 fixes the issue.
I'm using Boost 1.87.0, Visual Studio 2019 (16.11.44) and Windows 11 (10.0.22631.4890).

laudrup:
It sounds like what you have done is correct and it ought to work. There are also tests to verify the functionality with different TLS versions so it's a bit strange.

The Windows implementation of TLS v3 is quite different from earlier TLS versions and even more complicated so it is very likely that you might have found an issue specific to TLS v3.

I have not implemented that specific part of the code though. Maybe @windowsair who was kind enough implement that can have a look at this?

windowsair:
I'd like to focus on the tls client first. It seems that the change cipher spec step is missing:

![Image](https://github.com/user-attachments/assets/5f8c80e4-9e90-411e-85c9-82f6590ca591)

windowsair:
Hi @laudrup ,

During the handshake phase, the received data packet was incorrect. On the second call to `InitializeSecurityContextA`, the `input_buffers_` did not meet expectations. It was expected to be the second TCP packet, but the actual dumped content was part of the first TCP packet.

![Image](https://github.com/user-attachments/assets/58afa86d-4938-4c4a-8311-27991bcb0b55)

laudrup:
@windowsair 
Thanks a lot for looking into this and even starting to work on a fix. It is highly appreciated as I don't really have the time to look into this myself currently.

windowsair:
Hi @mkentie , 

Could you try this pr #92  ? Currently, only synchronous methods are available. I've already tested it on the echo_client/server and https_client examples.

mkentie:
I'm on vacation, but I'll take a look once I'm back in a week!

mkentie:
[...]

PR #92, opened 2025-03-13
Fix tlsv1.3 handshake issue

windowsair:
#91 
That’s only part of it. Besides, we need to handle the `SEC_I_RENEGOTIATE` situation. 
> See: https://learn.microsoft.com/en-us/windows/win32/secauthn/recognizing-a-request-to-renegotiate-a-connection

I’m not sure when I’ll be able to finish them.


laudrup:
@windowsair
Looks good. Thanks a lot for you help with this and for your efforts in supporting this project in general.

PR #93, opened 2026-02-18
Enable TPM-backed key support for client authentication via boost-wintls

MatthiasVonLaer:
This change enables the use of a TPM-backed key for client authentication. Since OpenSSL does not support TPM-backed keys without additional extensions, boost-wintls is a suitable alternative.
To make this work, I added the following line:
`CRYPT_ACQUIRE_ALLOW_NCRYPT_KEY_FLAG`
This allows acquisition of private keys managed by CNG providers, including those backed by TPM.
As I am not deeply familiar with Schannel and its edge cases, please review for any potential side effects or security implications.

### DIFF
--- a/include/wintls/detail/context_certificates.hpp
+++ b/include/wintls/detail/context_certificates.hpp
@@ -87,7 +87,7 @@ public:
     DWORD unused_1;
     BOOL unused_2;
     if (!CryptAcquireCertificatePrivateKey(cert,
-                                           CRYPT_ACQUIRE_COMPARE_KEY_FLAG,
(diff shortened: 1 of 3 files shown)


laudrup:
Thanks a lot for adding a unit test. Looks promising.

I don't have much time currently but I'll have a look at it sometime soon. I can see that the CI jobs fail and it does look like its your new job that causes the failure so maybe you could have a look at that?

The reason they all time out instead of fail is that I added an action for setting up an SSH connection in case of failure so I could investigate CI problems. Whether that was a good idea is probably up for debate :-)

laudrup:
Thanks a lot.

There seems to be a problem with the MinGW build but that doesn't seem like it's related to your changes at all so I'll have to look into that at some point.

I'll merge this as it is now and consider making a new release soonish.

Thanks once again.

MatthiasVonLaer:
Thank you as well!

